PR #145, opened 2025-01-21
AI Agent v0

mvanniasingheTT:
# Description 

This PR adds support for the search agent using the Llama 70B Instruct model. Using the ChatUI, a user is able to switch between using their standard LLM of choice and using the search agent to access real time information from the web. 

# Instructions to Setup Agent 

## Build Docker Image 

Once you have done the typical setups with the weights and persistent volume, follow the below: 

Visit https://tavily.com/  to generate an API key to use the search tool.  Update the `tt-studio/app/.env.default` with the API key

```bash
TAVILY_API_KEY=<your-tavily-api-key>
```

```bash
docker pull ghcr.io/tenstorrent/tt-studio/agent_image:v1
cd tt-studio/app
docker compose up --build
```

Refer to the below video for an example of what the agent can be used to do: 

https://github.com/user-attachments/assets/d06486a4-adce-4823-8a66-57d506cbabd4

The overall flow of tt-studio with the agent is shown below: 
Screenshot 2025-01-24 at 3 51 43 PM

The details of how the agent and the llm work together are depicted below: 
Screenshot 2025-01-24 at 3 51 04 PM

To Do:
- [x] Add agent image to GHCR (?)
- [ ] Thread id of agent should by dynamic based on which chat being used in ChatUI


app/api/model_control/model_utils.py
```diff
@@ -48,6 +49,49 @@ def health_check(url, json_data, timeout=5):
         logger.error(f"Health check failed: {str(e)}")
         return False, str(e)
 
+def stream_response_from_agent_api(url, json_data):
+    try:
+        new_json_data = {}
+        # TODO: thread_id should be dynamic once there is support for multiple different chats with no shared context
```
milank94, line 55: Does this mean that only 1 active chat can use the agent application?

mvanniasingheTT: multiple active chats can use the agent application, but they were all using the same agent running on the same thread. I made a fix so that each chat uses an agent running on its on thread. This is mainly for memory/stateful chat purposes, but since it chat has its own state it doesn't really change too much. The fix is here: https://github.com/tenstorrent/tt-studio/pull/168/commits/d33c3b518cdaf83b6ce831e59d29b428c88eb4f6

PR #163, opened 2025-01-27
Ben/sd1.4

bgoelTT:
This PR integrates the stable-diffusion-1.4 inference server into TT-Studio. The features include:
- launching & destroying the server from the studio UI
- a new Image Generation page
  - agnostic to the model being used (allows for future model integration such as a better stable-diffusion version, i.e., sd-3.5)
  - looks and functions similar to the ChatUI where the user submits prompts to the server and the UI "replies" back to the user with their generated image

Some future improvements:
- adding a stateful chat system similar to ChatUI (currently if you navigate to a different page all the state information is destroyed)
- Swapping the current initial sample image UI with something similar to ChatUI (example sample prompts to show the user how to generate the best images)
- enable queuing of multiple image generation requests
  - the server can already handle this, needs frontend support


bgoelTT:
Superceded by #193

Issue #249, opened 2025-03-24
RAG Feature Enhancements for TT-Studio and AI Playground


anirudTT:
This issue tracks all new features and improvements needed for the RAG system across both TT-Studio and AI Playground.
Each sub-issue represents a specific enhancement aimed at improving user experience, reliability, or maintainability.

anirudTT:
This issue is closed and is tested on a dev branch- #238

PR #255, opened 2025-03-25
Anirud/rag data using cookies

anirudTT:
### Changelog
- #249 

> - Each user’s queries are filtered through only the documents and data they are authorized to access.
> - Administrators can designate visibility or access rights for certain documents, ensuring compliance with privacy or organizational data policies.
> - When a user interacts with the AI Playground, the RAG system retrieves relevant information from only those sources that match the user’s permissions or context.
> 

Screenshot 2025-03-27 at 10 26 42 AM

Screenshot 2025-03-27 at 10 26 56 AM

Screenshot 2025-03-27 at 10 27 27 AM

Screenshot 2025-03-27 at 10 28 39 AM

### DIFF
--- a/app/.env.default
+++ b/app/.env.default
@@ -7,9 +7,18 @@ VLLM_LLAMA31_ENV_FILE=""
 JWT_SECRET=test-secret-456
 DJANGO_SECRET_KEY=django-insecure-default
 TAVILY_API_KEY=tvly-xxx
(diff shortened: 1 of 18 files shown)


anirudTT:
Uploading pdf within InputArea of chaui will result in this dialog box which warns users that they cant do it this way can should instead use the inbuilt rag
Screenshot 2025-03-27 at 11 00 47 AM

PR #303, opened 2025-04-29
Yolo v4 and Whisper Fixes

sbennettTT:
-Bounding boxes fixed
-Metrics light dark mode fixed
-Metrics icon and display on webcam and image fixed

### DIFF
--- a/app/frontend/src/components/object_detection/ObjectDetectionComponent.tsx
+++ b/app/frontend/src/components/object_detection/ObjectDetectionComponent.tsx
@@ -4,56 +4,33 @@ import { useLocation } from "react-router-dom";
 import { customToast } from "../CustomToaster";
 import React, { useState, useCallback, useRef, useEffect } from "react";
 import { Card } from "../ui/card";
-import { motion, AnimatePresence } from "framer-motion";
+import { Tabs, TabsContent, TabsList, TabsTrigger } from "../ui/tabs";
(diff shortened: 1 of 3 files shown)


anirudTT (changes requested):
Requested modifications:

- [ ] Restore object highlighting functionality: Re-enable the hover interactions that highlight detected objects on the live webcam feed when selected in the results list. This feature exists in the original code and simply needs to be reactivated. It provides essential visual connection between listed detections and their actual location. This needs to be re-enabled on webcam and file upload view.

- [ ] Reposition stop button: Move this control outside the webcam viewing area to prevent obstruction and improve accessibility. 

- [ ] Keep animation effects: Retain the existing motion/appear animations from the original code. These subtle effects provide valuable visual feedback and contribute to the polished feel of the interface without causing performance issues.

All three features existed in the original implementation and I think we need restore those back. Keeping the UX/UI of the webcam page + file upload the same is excellent!

anirudTT:
Screenshot 2025-04-29 at 8 53 38 PM

> This is the object highlighting functionality

anirudTT:
@sbennettTT   I also think we should keep the original positions of keeping the table and the file/webcam side by side but open to feedback

sbennettTT:
Let's discuss tomorrow

sbennettTT:
Will do!

PR #304, opened 2025-04-29
Chat UI fixes

sbennettTT:
- Autoscroll
- Autoscroll Breakaway
- Various UI fixes
- Proper implementation of theme
- like and dislike buttons while providing no feedback, do something for a better UX


app/frontend/src/components/chatui/MessageActions.tsx
anirudTT, line 1: I suggest we omit all the changes in this file and revert to back to original.

anirudTT (changes requested):
cool the scroll + stats light mode fix are perfect
let's omit changes in MessageActions and revert back to original. 
I created this issue #305  while testing - if u wanna fix that too, would be nice but not required for this PR.

sbennettTT:
I'll fix it :)

PR #308, opened 2025-05-01
Yolov4-Whisper final fixes and new PR

sbennettTT:
### Changelog
- #309 
Fixed Webcam Bounding issues
Will move Switch camera into a new PR


anirudTT (changes requested):
Great Changes 💯 
Few minor corrections

- [ ] Can we display detection results with just the confidence scores and detected objects, while hiding the rest of the data? This approach was included in the original code and would help prevent user confusion when they're viewing these essential elements.
- [ ] Can we revert the uploaded file layout area to its previous design? I'd like it to be less visible by default, but show better visibility when users hover over it. This would improve the overall layout and user experience.

@sbennettTT : Lmk if you have any thoughts or concerns on the feedback. 

----
Attaching some screengrabs from the original code
----
Screenshot 2025-05-01 at 10 46 30 AM
Screenshot 2025-05-01 at 10 46 46 AM

sbennettTT:
as for the second request, the 2 tabs are very similar now, and I think it looks good. try a vertical image, the bounding boxes are off, and the image is cut short. I think my fixes work better, but still not sure what you meant. let's discuss when you have a chance?

sbennettTT:
Completed the changes with filename and remove image colouration over hover

sbennettTT:
the bounding box is still to high on webcam, without changing anything else, fix that

anirudTT:
Think we need to update branch + add updated PR changelog .

PR #313, opened 2025-05-13
Anirud/whisper model local bring up

anirudTT:


anirudTT:
- [x] tested TT-Studio being able to deploy and user being able to use Whisper model
- [x] Tested that AI Playground version also will work on the same fork

PR #441, opened 2025-07-17
Shane/lightdark fixes

sbennettTT:
light dark cherry picks


sbennettTT (commented):
copilot suggestions incorporated where necessary

PR #444, opened 2025-07-18
Fix: unreadable error lable in light mode

### LINKED ISSUE #443: Error label text not visible in light mode
**Describe the bug**
The "Error" label appears with white text on a light background when using the app in light mode, which makes it very hard to read. In dark mode, the same label looks fine since the white text stands out properly against the dark background.

**To Reproduce**
1. Go to 'Home Page'
2. Open TT Studio in light mode
3. Trigger an error state (like not connected with backend properly)

**Expected behavior**
The text inside the error badge should switch to black (or a dark color) in light mode so that it stays readable. Dark mode can continue using white text as it currently does

**Screenshots**

Image

AnkittChauhan:
## What does this PR do ?
Fixes the "Error" label text color in light mode for better readability 

Fixes #443  

Screenshot from 2025-07-18 19-25-46

### DIFF
--- a/app/frontend/src/components/Footer.tsx
+++ b/app/frontend/src/components/Footer.tsx
@@ -156,7 +156,7 @@ const Footer: React.FC<FooterProps> = ({ className }) => {
                   ? "destructive"
                   : "default"
             }
-            className="text-xs"
+            className={`text-xs ${textColor}`}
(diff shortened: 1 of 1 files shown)


anirudTT (approved):
LGTM 👍🏼

PR #468, opened 2025-07-23
shane/remove-auto-browser-load

sbennettTT:
- #467


anirudTT (changes requested):
We need to revert this change in the `api.py`. 

Screenshot 2025-07-26 at 10 39 07 PM

PR #620, opened 2026-02-11
fix-artifact-switching-features

anirudTT:
feat: Improve TT Inference Server artifact management and UX                                                                                      
                                                                                                                                                    
  ## What Changed                                                                                                                                   
                                                                                                                                                    
  ### 🔐 Sudo Fallback for Permission Issues                                                                                                        
  - Automatically uses sudo when artifact cleanup fails due to permissions                                                                          
  - Asks for user confirmation before running sudo commands                                                                                         
  - No more manual cleanup needed when switching artifacts

  ### ✅ Better Artifact Validation
  - Validates artifact structure after download (workflows/, utils.py)
  - Clear error messages if download is corrupted
  - Consistent checks across all download types

  ### 📄 Redesigned artifact-info.txt
  Shows at a glance what you're using: see screenshot below .

  ### 🔄 New Flag: `--reconfigure-inference-server`
  - First run: Prompts for branch or version choice
  - Later runs: Uses saved config silently (no prompts!)
  - To change: `python run.py --reconfigure-inference-server`

  ## Why This Matters
  - **No more permission errors** blocking artifact switches
  - **Faster repeat runs** - no unnecessary prompts
  - **Clear visibility** into what artifact you're using

  ## Usage
  ```bash
  # Normal run (silent if config exists)
  python run.py

  # Change artifact
  python run.py --reconfigure-inference-server

### DIFF
--- a/run.py
+++ b/run.py
@@ -1469,31 +1469,63 @@ def configure_inference_server_artifact(dev_mode=False, easy_mode=False, force_r
         if current_branch:
             # Clear branch if switching to release
             write_env_var("TT_INFERENCE_ARTIFACT_BRANCH", "", quote_value=False)
-        
+
         # Always prompt for version when user chooses option 1
         default_version = "latest"
         if current_version and current_version != "latest":
             default_version = current_version
-        
+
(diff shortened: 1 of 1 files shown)


rfatimaTT (approved):
lgtm

PR #710, opened 2026-04-07
Improve model deletion flow

rfatimaTT:
### Streaming Model Deletion with Live Logs
Replaces the synchronous model deletion flow (toast notifications + hardcoded delays) with a real-time SSE-streamed dialog that shows exactly what the backend is doing. 

#### Backend
- New async SSE endpoint GET /docker-api/stop/stream/<container_id>/ (StopStreamView) that performs stop → remove → - board reset sequentially, streaming each step's logs in real time
- Uses asyncio.create_subprocess_exec for tt-smi so output streams line-by-line instead of all at once
- Handles 404 on stop/remove gracefully (container already gone = success, not failure)
- Strips ANSI escape codes from tt-smi output
#### Frontend
- New useDeleteStream hook that connects to the SSE endpoint and buckets logs per step (deleting / resetting)
- Redesigned DeleteModelDialog with per-step inline log panels — container logs under step 1, tt-smi output under step 2
- Auto-closes dialog 1.5s after successful completion and triggers model list + hardware refresh
- Removed all toast notifications from the deletion flow
- Cleaned up deleteModel() API function (stripped toasts; still used by reset flows)

UPDATE
- remove stale models with status "unknown" without refresh and show "Your Model may have stopped unexpectedly or the board was reset"

### DIFF
--- a/app/backend/docker_control/urls.py
+++ b/app/backend/docker_control/urls.py
@@ -7,6 +7,7 @@ from django.urls import path
 from . import views
 from .views import (
     StopView,
+    StopStreamView,
     ContainersView,
     StatusView,
     ChipStatusView,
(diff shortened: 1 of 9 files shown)


rfatimaTT:
image

rfatimaTT:
self-deleted model


image

rfatimaTT:
stale model 


image

anirudTT (approved):
LGTM

PR #717, opened 2026-04-14
Added Qwen-only host-volume preload support in the inference FastAPI layer

anirudTT:
- For Qwen3-32B, FastAPI now does a _better check_ for preloaded host-volume layout under `~/data/tt-cache` before adding --host-volume.
- If the expected Qwen weights/cache layout is present, _uses_ host-volume mode.
- If not, it _falls back_ to the existing baseline startup automatically, which is without the above flag.
- The model list that can use this flag is now configurable via `inference-api/host_volume_models.json`
- Added logging so you can see which Qwen path was checked and whether host-volume or baseline mode was used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,6 +71,7 @@ CLAUDE.md
 docs/RAG_PRODUCTIONIZATION_PLAN.md
 docs/DOCKER_CONTROL_SERVICE_PLAN.md
 !app/backend/shared_config/models_from_inference_server.json
+!inference-api/host_volume_models.json
 
 request-venv/*
 app/.env-old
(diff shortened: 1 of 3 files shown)


anirudTT:
@rfatimaTT Can you check whether this works for your user on the QB2 VM we have?

For Qwen, as long as the expected directory is present, it should use the host-volume flag. For something like Llama 8B, it should not use that flag, lets make sure to monitor the logs to check for this.

anirudTT:
- [x] Tested that 8b models deploys without adding host volume flag
- [x] Tested on n300 for any random model 
- [x] Tested that qwen 32b models deploys with the host flag being added when host cache is present . 
Screenshot from 2026-04-15 04-08-11
Screenshot from 2026-04-15 04-07-38

anirudTT:
merging by bypassing approvals.

PR #718, opened 2026-04-14
Anirud/stale deployment fixes

anirudTT:
- Fixes https://tenstorrent.atlassian.net/browse/DEVSTACK-68

### Summary

This change fixes a stale deployment-state bug where the deploy flow could show All Devices Occupied even when no models were actually running.

ChipSlotAllocator now reconciles saved ModelDeployment records against the live Docker container snapshot before reporting chip occupancy. Stale running records are marked terminal, stale starting placeholder records are cleaned up after a grace period, and valid placeholder records are promoted to running when they match a live container.


anirudTT:
For example : 

Screenshot 2026-04-14 at 3 10 21 PM
Screenshot 2026-04-14 at 3 10 30 PM

rfatimaTT (approved):

PR #728, opened 2026-04-22
Rumeza/search agent

rfatimaTT:
- cleaned up search agent 
   - user can now see search process, selected links that the agent used in the clean format 
- improved metrics output for search-agent related inputs
- currently best suited to be used by Llama instruct model 
image


anirudTT (changes requested):
Image
Image




### Some comments / suggestions 

* [x] Show a clear indicator when the search agent is active so the user knows it’s running
* [x] If tool calling is not enabled, hide the search agent feature entirely
* [x] The sources UI looks good, but sometimes sources also get listed in the message body. We may need to tighten the instructions so source rendering is more consistent
* [x] We currently have two different UIs for showing sources. We should standardize on one. I’m leaning toward the top-level sources UI, but we should think through how that interacts with thinking models, since thinking tokens could conflict with that layout
* [x] We may need stronger error handling as well. I was able to break the agent in a few cases

Image

anirudTT:
**Testing plan**

* [ ] Deploy LLM models on the QB2.
* [ ] Trigger the Search Agent from the input area after setting up the Tavily API key from the Tavily website as an environment variable.
TAVILY_API_KEY="tavily-api-key-not-configured"

app/backend/docker_control/views.py
```diff
@@ -338,6 +338,16 @@ def post(self, request, *args, **kwargs):
             BASE_SERVICE_PORT = 7000
             service_port = BASE_SERVICE_PORT + device_id
 
+            # Build vllm_override_args when tool calling is requested
+            vllm_override_args = None
+            if serializer.validated_data.get("enable_tool_calling"):
+                import json as _json
+                vllm_override_args = _json.dumps({
+                    "enable-auto-tool-choice": True,
```
anirudTT, line 346: Could we avoid hardcoding `tool-call-parser: llama3_json` here and pull the parser from the model spec/config instead?
 This would let Qwen-based models use the Hermes parser while keeping `llama3_json` as the default for Llama models.

anirudTT (changes requested):

rfatimaTT:
image
image

jashansinghTT:
This will be merged with a unified PR through jashan/canvas-on-dev

PR #750, opened 2026-05-07
Move first-run setup and runtime secrets into the UI

jashansinghTT:
Moves first-run setup and runtime secrets (HF token, TTS, Tavily, JWT) out of the CLI and into the UI: a Welcome stepper on first run and a Settings dialog afterwards, both backed by `GET/POST /settings/` plus an HF access check endpoint.

Secrets are stored in a single dotenv file in the persistent volume, `backend_volume/user_config.env` (0600, atomic writes, values sanitized against key injection). An existing `user_config.json` from earlier builds is migrated automatically and deleted. Consumers (backend, agent, inference-api) read it fresh on each use, so Settings changes apply without redeploying; env vars in the root `.env` remain a fallback.

`run.py` no longer prompts for these secrets, and `--cleanup` resets the Welcome flag by dropping `SETUP_COMPLETE` from the file while keeping saved tokens. `dev` is merged in; deploys keep the HF access pre-check but no longer stop same-model deployments (multiple instances stay allowed, #990).

Tests: `shared_config/test_user_config.py` covers round-trip, permissions, sanitization, and migration; existing cleanup and docker_control suites pass.


anirudTT:
- [ ] expose the following env variables, so all of these should be configurable at runtime . 
     - [ ] hf token 
     - [ ] JWT : pre configured for the user 
     - [ ] TTS: TTS_API_KEY
     - [ ] TT_INFERENCE_ARTIFACT_VERSION or TT_INFERENCE_ARTIFACT_BRANCH ==> better explain this to them . ( we will expose this , but not allow them to change this via the UI , right now on this PR , this would be a separate issue.
     - [ ] travily api key for search  


- [ ] remove hf token ask / access on the run.py CLI , shift all this to a new home ui page , that user sees on first run. 
           All these checks move into the UI now. 
                HF token access check:
                   ✅ Llama 3.1: access confirmed
                   ✅ Llama 3.3: access confirmed
                   ✅ Qwen3-32B: access confirmed

anirudTT:
Screenshot 2026-05-08 at 12 21 52 PM
move this as well into the new settings page

anirudTT:
Daily:
- [x] @anirudTT + @rnabeelTT to test this after rc.

anirudTT (changes requested):
## Suggestions

- **Keep TTS key as a default secret** — let's keep it as a default value for now rather than dropping it, we can revisit later.
- **Auto-run the HF access check** — it's not obvious you need to click "Run check" on mount. Trigger automatically and repurpose the button as "Re-check."
- **Show a "Saved" badge on secret fields** — the placeholder is easy to miss. A green check + "Saved" badge when a secret is already set would be much clearer.
- **Show the gear icon on the Done screen** — the copy references it but doesn't show it. Render it inline, ideally as a button that opens Settings directly.

anirudTT:
We also don't also want to do this when a user runs `--cleanup`, it should store their secrets 

Screenshot 2026-05-19 at 5 13 19 PM

anirudTT:
@jashansinghTT can u resolve the conflicts

anirudTT:
## Review notes: Settings / Secrets UI

Ran a deep pass over the secrets-in-UI flow (Welcome + Settings dialog + backend). Grouping into **blocking bugs**, the **"applies immediately" accuracy gap**, and **UX polish**. None of this needs to be solved in this comment — capturing it so it's tracked.

### Blocking bugs

- [ ] **`NameError` crashes local image/video generation.** The module global `TTS_API_KEY` was removed and TTS/speech call sites moved to `get_tts_api_key()`, but four bare `TTS_API_KEY` references remain in `app/backend/model_control/views.py` (lines ~631, 705, 788, 818 — Image/Video generation views). Any local FLUX/SD/Wan request raises `NameError: name 'TTS_API_KEY' is not defined`. Should be `get_tts_api_key() or ""` like the other sites.
- [ ] **UI-set HF token can miss the download path.** `app/backend/model_control/download_progress.py:91` reads `os.getenv("HF_TOKEN", "")` directly, bypassing the `get_hf_token()` resolver. A token saved only through the UI (not exported to the backend process env) may not reach weight downloads.

### Make it clear which secret powers what, and when it applies

The dialog presents the secrets as a flat list and says all changes apply immediately. Two things to make explicit:

**What each secret is for (group the UI by this):**

| Secret | Powers | Applies live? |
|---|---|---|
| `hf_token` | **Both** LLM *and* media models — gated weight downloads | Next deploy only; a model already running keeps its old token until redeployed |
| `jwt_secret` | Auth between backend ↔ deployed model containers (LLM serving) | Effectively always live — auto-managed, stable, not user-editable |
| `tts_api_key` | Media / voice models (TTS inference) | Yes — resolved per request |
| `tavily_api_key` | Search agent | Yes — read per search call |

By model type: **LLM → HF + JWT**, **media (image/video/voice) → HF + TTS**.

**"Applies immediately" is only wrong for one field:** `SettingsView.post` returns `requires_redeploy: false` unconditionally, but `hf_token` is injected into each model container at deploy time (`tt_inference_client.py:134`), so a model deployed before the token changed must be redeployed. Suggest making `requires_redeploy` field-dependent and messaging only the HF-token case as "affects new deployments; redeploy existing models."

### UX (from testing the running app)

- [ ] **No way to see a token you just entered.** All fields are `type="password"` with no reveal toggle, so you can't verify a paste. (Note: the on-disk `user_config.env` is intentionally `0600` root-owned and unreadable from the host — the UI is the only place to confirm state, so this matters.) A reveal-while-typing toggle would help; saved values can stay masked.
- [ ] **Set vs unset, and source, are unclear.** `SettingField.set`/`source` are returned by the API but the UI ignores `source`, so a value inherited from `.env` looks identical to one saved via the UI. Suggest distinct states: "Saved via UI" / "From .env (fallback)" / "Not set".
- [ ] **HF access check placement.** `<HfAccessCheck />` sits at the bottom of the dialog and receives no `token` prop, so it validates the *stored* token, not what was just typed. Move it into a collapsible directly under the HF token field and pass the typed value.
- [ ] **Read-only inference artifact needs a "how to change" hint.** The box says editing is disabled but not how to change it. It's `python run.py --reconfigure-inference-server` (or edit `.env` `TT_INFERENCE_ARTIFACT_VERSION`/`_BRANCH` and redeploy). Surface that command in the helper text.
- [ ] **Blank save reports success.** Submitting an all-blank form still toasts "Settings saved" and closes; the `updated`/`requires_redeploy` response fields are ignored.
- [ ] **Welcome HF-check dead-ends on network error.** `HfAccessCheck`'s catch branch never calls `onChecked`, leaving Continue disabled with no way forward on a transient failure.

### Lower priority

- Temp-file permission race in `user_config.py` (`.env.tmp` written before `chmod 0600`).
- `_mask` reveals up to 8 chars of short secrets.
- `SettingsView.get` has a write side-effect (JWT auto-gen persists on first GET) and can mislabel the jwt `source` as `env`.
- `--stop` (the flag that resets the Welcome flag while keeping tokens; `--purge-all` intentionally wipes everything) falls back to deleting the whole `user_config.env` — wiping saved secrets — when the file is root-owned, despite the "preserve secrets" intent.

### Answered while reviewing
- `tts_api_key` **is** editable via Settings.
- Tavily field layout looks good.
- The `NoPermissions` error opening `user_config.env` in an editor is **expected** — it's a 0600 root-owned secrets file by design.

jashansinghTT:
Thanks for the thorough pass @anirudTT — really helpful. Pushed fixes for basically all of it:

- Both blockers gone: the `TTS_API_KEY` NameError (image/video views resolve it at request time now) and the HF-token download path (`download_progress` uses `get_hf_token()` instead of `os.getenv`).
- `requires_redeploy` is now field-aware — only true when the HF token changes, and the toast tells you to redeploy running models. Blank Save is a no-op instead of a scary 400.
- HF access check moved up under the token field and tests the value youre typing, not just the stored one; the network-error dead-end is fixed too.
- Added a reveal toggle on the secret inputs, and the badges now say "Saved via UI" / "From .env (fallback)" / "Not set" so source is obvious.
- Artifact box now shows the `--reconfigure-inference-server` hint.
- Backend nits: short secrets fully masked, jwt `source` label fixed on first GET, `user_config` temp file created 0600 (no more race), and the deploy pre-check falls back to `model_index.json` for FLUX/Wan.
- `--stop` cleanup now does an atomic 0600 rewrite that keeps your secrets, strips the flag in-place for legacy JSON, and tries a sudo in-place strip before ever deleting a root-owned file.

Added a few tests for the redeploy flag, no-op save, masking, and the diffusers fallback. Left the JWT auto-persist-on-GET as-is (label is correct now). Lmk if I missed anything!

anirudTT:
- [ ] @jashansinghTT : will test non dev / prod mode

anirudTT (approved):
lgtm , great work here @jashansinghTT

PR #755, opened 2026-05-11
Rc v2.6.0

anirudTT:
# TT-Studio v2.6.0-rc — Test Plan

## Setup

- **Host:** `p300x2`
- Start from a fresh install / clean Docker state before testing

---

## Part 1 — Core Deployment

Deploy each model and confirm the inference server starts with no errors.

| # | Model | Flags | Pass? |
|---|---|---|---|
| 1 | Qwen3-32B | `--host-volume` | [ ✅] |
| 2 | Llama-70B | default | [ ] |
| 3 | Llama-8B | default | [✅ ] |
| 4 | TTS single-model deploy | default | [N/A] |
| 5 | Whisper / Speech Recognition single-model deploy | default | [N/A] |

### Validation

- For chat models 1–3:
  - [x] Send a message
  - [x] Confirm a streaming response is returned

- For TTS / Whisper:
  - [x] Confirm the container is healthy
  - [x] Confirm the respective demo UI works

> **Note:** Only LLM models can be individually deployed using the Single Deployment interface for this release. Most users are expected to access them through the Solution Deployment Pipeline in Part 2. Fixing the UI/UX for the single-model path is out of scope for this RC and will be addressed in the next release.

---

## Solution Deployment Pipeline (#739)

_Primary user path for TTS + Whisper / Talk to the Box pipeline._

- [x] Stepper shows **Solution** vs **Single/Multi** choice
- [x] Solution path opens the **Voice Agent** step
- [x] Voice Agent solution deploys the full stack end-to-end:
  - [x] LLM
  - [x] TTS
  - [x] Whisper
- [x] Deployment progress bar updates while containers start
- [x] All sub-deployments land in `running` state

---

## Multi-chip Deployment (#741)

- [x] Deploy a model on a specific card pair, for example devices `0,1`
- [x] ModelsTable shows the correct device IDs
- [x] Single-chip deployment still works

> **Note:** We have temporarily disabled manually configuring these options.
---

## Stale Deployment Cleanup (#718, #739)

- [x] Run `docker kill` on a running container
- [x] Confirm the stale deployment record auto-cleans within ~1 minute
- [x] Start a new deployment immediately after
- [x] Confirm the new deployment is **not blocked**

---

## Bug Report (#732)

- [x] Click **Bug Report** in the navbar
- [x] Submit a report
- [x] Confirm the report includes:
  - [x] Logs
  - [x] Current models snapshot
  - [x] Diagnostics reference ID in the issue body

---

## Chat Suggestions (#722)

[...]

PR #556, opened 2025-12-02
fix: firefox mic access error for SR modal

AnkittChauhan:
## What does this PR do ?
Fixes the `mic/speech-to-text` issue on Firefox browsers by removing the hardcoded sampleRate and allowing the browser to use its default audio configuration

Fixes #317 

https://github.com/user-attachments/assets/a90b718e-2e8c-442f-b180-69ede37582c3

PR #767, opened 2026-05-13
fix: firefox mic sample-rate mismatch on STT + voice-agent

jashansinghTT:
## Summary

- Fixes the Firefox-only `DOMException: AudioContext.createMediaStreamSource: Connecting AudioNodes from AudioContexts with different sample-rate is currently not supported.` reported in https://github.com/tenstorrent/tt-studio/pull/755#issuecomment-4443314170.
- Mirrors #556's approach (which targets `speechToText/AudioRecorderWithVisualizer.tsx`) and **extends the same fix to `voiceAgent/AudioRecorderWithVisualizer.tsx`** — that file's line 91 is the exact location in the user-reported stack trace, and it was added as part of #755's voice-agent feature so #556 doesn't cover it.

## Why this fixes it

Both files hardcoded `sampleRate: 16_000` on the `getUserMedia` constraint AND the `AudioContext` constructor. Firefox returns streams at the device's native rate (typically 48 kHz) regardless of the requested constraint. The forced-16 kHz `AudioContext` then rejects `createMediaStreamSource` because the rates don't match. Dropping the explicit `sampleRate` lets the browser pick a consistent rate for both sides.

Net diff: 4 insertions, 7 deletions across 2 files. No behavior change in Chromium.

## Test plan

- [x] `npm run type-check` passes
- [x] `npm run lint` — same 15 pre-existing problems on these files before and after; no new lint issues introduced
- [x] Playwright-driven Firefox 150 (headless) loads `/speech-to-text` and `/voice-agent`, clicks the mic, and the recorder flow completes without `pageerror` or mic-related `console.error`
- [ ] **Manual verification on a real Firefox with a real mic** — recommended; modern Firefox 150 auto-resamples internally so the original DOMException is no longer reproducible headlessly, but the reporter's environment (older Firefox build / different audio backend) is what surfaces the bug. Click the mic on both `/voice-agent` and `/speech-to-text`, confirm the DevTools console is clean of any `createMediaStreamSource` error.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/frontend/src/components/speechToText/AudioRecorderWithVisualizer.tsx
+++ b/app/frontend/src/components/speechToText/AudioRecorderWithVisualizer.tsx
@@ -77,8 +77,6 @@ export const AudioRecorderWithVisualizer = ({
   const animationRef = useRef<number | null>(null);
   const audioRef = useRef<HTMLAudioElement>(null);
 
-  const sampleRate = 16_000; // 16kHz sample rate
-
   function cleanupAllResources() {
     // Stop recording if in progress
     if (
@@ -150,7 +148,7 @@ export const AudioRecorderWithVisualizer = ({
     if (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
       navigator.mediaDevices
         .getUserMedia({
-          audio: { sampleRate },
+          audio: true,
         })
         .then((stream) => {
           setIsRecording(true);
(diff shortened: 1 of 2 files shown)


anirudTT (changes requested):
this is the wrong branch to open it off to. 
you need the PR off to `dev`

anirudTT:
also I think you forked the repo and adding this PR , is there a reason why?

jashansinghTT:
Superseded by #790 — same commit (`851dcef`), but sourced from the `origin` branch `tenstorrent/tt-studio:jashan/firefox-fix` (matching the team's normal workflow) and targeting `dev` per the earlier review. Closing here; please continue review on #790.

PR #774, opened 2026-05-14
Surfaced error on Deployed Models page in case deployment fails unexpectedly

### LINKED ISSUE #773: Warn user to reset board after model deploy failure due to engine/core issues
**Is your feature request related to a problem? Please describe.**
Currently, when a user's model deployment fails due to engine core or similar internal errors, there is no direct warning to the user about what went wrong or clear next steps. This can cause confusion when attempting to deploy the voice agent pipeline again.

**Describe the solution you'd like**
Warn the user if their model deployment failed due to an engine core or other internal hardware/software issues. The system should inform the user that they need to perform a complete board reset before attempting to redeploy the voice agent pipeline. This warning should be clear, actionable, and visible immediately upon deployment failure.

**Describe alternatives you've considered**
- Not resetting the board and retrying deployment (often unsuccessful)
- Only providing a generic error message (not helpful)

**Additional context**
This improvement will reduce deployment failures and improve user guidance in recovery situations.

rnabeelTT:
Models that die unexpectedly due to an inference server error are now shown on the Deployed Models page with an option to view logs or remove them from the page. Resolves issue #773. 

Logs for this issue can be found on: https://gist.github.com/rnabeelTT/66131a98d0a904132790be959d5fdce1

Screenshot 2026-05-13 at 5 28 36 PM

### DIFF
--- a/app/frontend/src/components/models/ModelsDeployedCard.tsx
+++ b/app/frontend/src/components/models/ModelsDeployedCard.tsx
@@ -33,6 +33,7 @@ import {
 } from "../../api/modelsDeployedApis";
 import type {
   ColumnVisibilityMap,
+  FailedDeploymentInfo,
   HealthStatus,
   ModelRow,
 } from "../../types/models";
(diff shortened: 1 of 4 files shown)


anirudTT (dismissed):
@rnabeelTT — I think we should add these recovery/debugging steps directly into the failure flow itself.

For example:

* direct link to deployment logs
* suggest deleting failed/stale deployments
* cleanup old containers
* TT-SMI reset option directly from TT-Studio if issues persist

That way users have a guided recovery path instead of repeatedly retrying deployments without understanding what failed or how to recover the system properly.

Also curious if the current implementation has been generalized across all deployment/runtime types, or if it only handles certain model flows today?

anirudTT (approved):
lgtm .

PR #794, opened 2026-05-15
Implement light persistent memory for user interactions in VoiceAgentApp

anirudTT:
- Added functionality to store short notes about the user across sessions using localStorage.
- Introduced methods to extract and manage user information from conversations, enhancing the assistant's contextual awareness.
- Updated the inference prompt to include memory notes, improving the conversational flow and user experience.


rnabeelTT (approved):
Works well, its usually able to store one fact from each conversation. It enhances the experience, but is a good beginning step towards voice agents with persistance.

anirudTT:
> Works well, its usually able to store one fact from each conversation. It enhances the experience, but is a good beginning step towards voice agents with persistance.

thanks for testing it.

PR #796, opened 2026-05-15
Enhance deployment logic to support full board allocation for Llama 3…

anirudTT:
deployment logic to support full board allocation for Llama 3.1 8B models

- Introduced a new `force_full_board` field in the DeploymentSerializer to allow explicit full board allocation.
- Updated DeployView to handle `force_full_board` requests, including validation and logging for Llama 3.1 8B models on P300Cx2 boards.
- Modified frontend SelectionSteps component to conditionally set `force_full_board` based on user selections, improving deployment flow for specific model configurations.


rnabeelTT (approved):
Utilizes all 4 chips for Llama deployments as expected.

PR #799, opened 2026-05-15
feat: reset TT Studio to fresh-clone state via --cleanup-all

jashansinghTT:
## Summary

Reworks `python run.py --cleanup-all` so it actually wipes every piece of generated TT Studio state and resets the working tree to fresh-clone condition, replacing the previous flag's partial behavior. Adds a non-interactive path, a browser-side wipe handshake, and unit-test coverage. Also extends basic `--cleanup` to re-trigger the Welcome flow on next bring-up.

### What it cleans up
- All `tt_studio_*` containers + the `tt_studio_network`
- **Deployment containers attached to `tt_studio_network`** (vLLM, YOLO, stable-diffusion, …) launched outside docker-compose by the backend via docker-control-service — `compose down -v` never sees them
- **Docker named volumes holding model weights** (`volume_id_*` — created by the inference-server side of the deployment pipeline; not declared in any compose file, so `compose down -v` doesn't touch them)
- **Dangling anonymous Docker volumes** (e.g. the frontend dev container's `/app/node_modules` anon volume from `docker-compose.dev-mode.yml`)
- `tt_studio_persistent_volume` (HF tokens, JWT secrets, deployment history, RAG vectors)
- `app/.env`, `.tt_studio_preferences.json`, `.tt_studio_easy_config.json`
- `.artifacts/` (downloaded inference servers, workflow logs, tarballs)
- `startup.log`, `fastapi.log`, `fastapi.pid`, `fastapi_logs/`, `docker-control-service.log/.pid`
- `output.wav`, `speech.wav` (TTS/STT scratch files)
- `inference-api/.venv`, `docker-control-service/.venv`
- Contents of `.workflow_venvs/` **except** the tracked `.venv_bootstrap_uv` symlink (a real QA pass surfaced that wiping the whole directory broke re-bootstrap; `_remove_directory_contents(..., preserve_names={".venv_bootstrap_uv"})` keeps it)
- Local Docker images: `ghcr.io/tenstorrent/tt-studio/*`, `ghcr.io/tenstorrent/tt-inference-server/*`, `ghcr.io/tenstorrent/tt-media-inference-server`, `chromadb/chroma`

### UX changes
- One upfront `[y/N]` confirmation with an inventory + total-size summary, instead of three separate prompts
- New `--yes` / `-y` flag for non-interactive use (CI, scripts)
- New pre-React sentinel handshake: `run.py` arms `app/frontend/public/.cleanup-pending` with a numeric-ms token; `app/frontend/src/main.tsx` reads it on next page load, clears IndexedDB (chat store) + localStorage, then renders React. Token is acked in localStorage so the wipe runs once.
- Factored existing teardown into `_cleanup_runtime`, so plain `--cleanup` keeps its original Docker-teardown-only behavior.

## Follow-up: Re-trigger Welcome on basic `--cleanup`

Basic `python run.py --cleanup` previously only tore down containers; the `setup_complete=true` flag in `user_config.json` stuck around, so the next bring-up skipped the Welcome flow even though the rest of state had just been wiped. Only `--cleanup-all` cleared it, but that also nuked model weights and the rest of the persistent volume.

- **`--cleanup` now resets just `setup_complete`** in `user_config.json` so the next bring-up walks back through Welcome. Saved secrets (HF token, TTS, Tavily) are preserved on the in-place rewrite path so the user only re-confirms them.
- **Sudo fallback via `_remove_path`** for the root-owned-by-Docker case. Docker writes `user_config.json` on the host volume as root, so a non-root user running `run.py --cleanup` hits `PermissionError` trying to rewrite it. On `PermissionError` we fall through `_remove_path(..., no_sudo=args.no_sudo)` and the backend regenerates a fresh `user_config.json` (with a new `jwt_secret` and no `setup_complete`) on the next bring-up.
- Scoped to basic `--cleanup`: `--cleanup-all` already wipes the entire `host_persistent_volume` (which contains `user_config.json`) as part of its inventory pass, so the reset block is intentionally skipped there.

(Moved here from #750, where it originally lived; it's a `--cleanup` UX change and belongs alongside the rest of the cleanup rework.)

## Follow-up: QA-surfaced leftovers

Three rounds of QA after the initial cleanup-all landed surfaced state that `compose down -v` doesn't reach. Each is fixed by a focused helper called from `cleanup_resources` in the order that respects volume/network/container dependencies.

**1. SDK-launched containers on `tt_studio_network`.** Deployment containers (vLLM, YOLO, stable-diffusion, …) and the agent/face-recognition direct-run containers attach to `tt_studio_network` but are not in docker-compose. `compose down -v` doesn't touch them, the follow-up `docker network rm` silently failed against active endpoints, and orphaned images couldn't be reclaimed because containers were still using them. New `_remove_tt_studio_network_containers` enumerates `docker ps -aq --filter network=tt_studio_network` and runs `docker rm -fv` (the `-v` is load-bearing — see anonymous-volume fix below). Runs first in `_cleanup_runtime` so the subsequent network removal and weight-directory deletion aren't blocked.

**2. Image filter was too narrow.** `_remove_local_tt_studio_images` filtered on a single hardcoded glob (`ghcr.io/tenstorrent/tt-studio/*`), which missed `chromadb/chroma` (third-party, declared in compose) and every inference-server image (`ghcr.io/tenstorrent/tt-inference-server/*`, `ghcr.io/tenstorrent/tt-media-inference-server`). Replaced with a `_CLEANUP_IMAGE_REFS` tuple iterating each pattern; image IDs are deduped and force-removed in one batch.

**3. Named volumes for model weights + dangling anon volumes.** `docker-compose.yml` declares no named volumes, so `compose down -v` doesn't help. The inference-server side creates volumes named `volume_{model_id}` (`volume_id_<impl>-<model>-v<ver>` — see `app/backend/shared_config/model_config.py:106-108`). These hold the actual weights for Llama 3.1-8B / 3.3-70B / Whisper / SpeechT5 / etc. New `_remove_tt_studio_model_volumes` filters on the `volume_id_` prefix with a Python-side `startswith` guard (docker's `name=` filter is substring match, not prefix). Separately, the frontend dev container declares an anonymous volume for `/app/node_modules`; without `-v` on `docker rm` those anon volumes orphaned. New `_prune_anonymous_volumes` is a defensive net — uses `docker volume prune --force` (without `--all`, so named volumes from other projects on the host are safe) and reports the count via a before/after `volume ls` diff.

### Files touched
- `run.py` — `--cleanup-all` rework, `--yes`, new helpers (`_format_bytes`, `_path_size`, `_remove_path`, `_remove_directory_contents`, `_remove_local_tt_studio_images`, `_remove_tt_studio_network_containers`, `_remove_tt_studio_model_volumes`, `_prune_anonymous_volumes`, `_write_browser_cleanup_sentinel`), `_CLEANUP_IMAGE_REFS` / `_CLEANUP_VOLUME_PREFIX` constants, `_cleanup_runtime` factor-out, basic-`--cleanup` Welcome-reset block
- `test_run_cleanup.py` (new) — 13 unit tests covering helpers, sentinel generation, abort path, destructive `--yes` path, `.venv_bootstrap_uv` preservation, ref/image filter coverage, volume prefix vs. substring guard, ordering of teardown steps
- `app/frontend/src/main.tsx` — pre-mount `applyCleanupSentinel()` handler
- `.gitignore` — adds `app/frontend/public/.cleanup-pending`

### Unit tests
- [x] `python3 -m unittest test_run_cleanup -v` — all 13 tests pass

### End-to-end on sjc2-qb2-9b22 (Blackhole quietbox)
- [x] Pre-state matching the QA-reported leftovers (from earlier failed cleanup-all attempts): 6 `volume_id_*` weight volumes (including Llama 3.3-70B-Instruct, ~140 GB total), 1 `tt-media-inference-server:qb2_launch-6900b0c` image (30 GB), 64 MB `tt_studio_persistent_volume/`, 124 MB `.artifacts/`
- [x] `python3 run.py --cleanup-all --yes` → inventory shown, "Stopping deployments... done", "Removing model volumes... done (6 volume(s))", "Pruning anon volumes... done", "Removing local images... done", "Removing host state... done (14/14 path(s))", "Arming browser wipe... done"
- [x] Post-cleanup assertions:
  - 0 containers on `tt_studio_network`
  - 0 `volume_id_*` volumes
  - 0 matching images across all four ref patterns
  - `docker volume ls` empty (confirms no orphaned anon volumes)
  - `tt_studio_persistent_volume/` absent (so `user_config.json` confirmed absent — verified separately via `find`)
  - `app/.env` absent
  - `.artifacts/` absent
  - `app/frontend/public/.cleanup-pending` written with a 13-digit numeric token
- [x] Fresh bring-up after cleanup: `python3 run.py` enters First-Time Setup with the T&C prompt — confirms `setup_complete` reset is exercised end-to-end
- [x] Targeted live verification of `_prune_anonymous_volumes` against two stale anonymous volumes left by a prior cleanup attempt (frontend dev container `node_modules`): helper returned `removed: 2`, `docker volume ls` now empty

### Basic `--cleanup` Welcome-reset
- [ ] Bring up the stack so the backend writes `tt_studio_persistent_volume/backend_volume/user_config.json` with `setup_complete: true`
- [ ] `python3 run.py --cleanup` (basic, not `--cleanup-all`) — assert `user_config.json` still exists, `setup_complete` is gone, and saved secrets (`hf_token`, `tts_api_key`, `tavily_api_key`) are preserved
- [ ] Re-bring up; UI redirects `/` → `/welcome`
- [ ] Root-owned file path: `python3 run.py --cleanup` on a `user_config.json` written by Docker (root-owned) falls back through `_remove_path` and removes the file; next bring-up regenerates it with no `setup_complete`
- [ ] `--cleanup-all --yes` does **not** print the basic-`--cleanup` Welcome-reset message (block is intentionally skipped in the full-cleanup path)

### Known gap
- The browser-side IndexedDB/localStorage wipe is verified via (a) unit test in `test_run_cleanup.py::test_write_browser_cleanup_sentinel_uses_numeric_token` for the token format and (b) live confirmation that the dev server serves the token at `/.cleanup-pending`. The React-side wipe itself was not driven through a real browser in this PR's QA — recommend a manual once-over on first review.


anirudTT (approved):
Minor nit: if you type anything other than `y/Y/yes/n/N/no` at the `Proceed? [y/N]` prompt it just aborts silently. Would be nicer to re-prompt the user instead of treating an unrecognized input as a cancel ; could open a small follow-up PR for this.

Lets also document this flag's usage in the readme in a new PR .
Tracking it here https://github.com/tenstorrent/tt-studio/issues/811

stisiTT (commented):
Verified findings from a multi-agent review pass. 4 issues — 2 critical, 2 high — each anchored to a specific file:line below. One earlier draft (about a 'broken tracked symlink' in `.workflow_venvs/`) was refuted on closer inspection: that file is in `.gitignore` and gets regenerated by `tt-inference-server/workflows/bootstrap_uv.py:54`, so the preservation logic is fine.

run.py
```diff
@@ -1663,13 +1663,379 @@ def display_welcome_banner():
...
+    """
+    sudo_prefix = ["sudo"] if not has_docker_access else []
+    try:
+        result = subprocess.run(
+            sudo_prefix + ["docker", "volume", "ls", "--filter",
+                           f"name={_CLEANUP_VOLUME_PREFIX}", "-q"],
+            capture_output=True, text=True, check=False,
+        )
+        # `--filter name=foo` is substring match; double-check the prefix in
+        # Python so we never delete an unrelated volume that happens to contain
```
stisiTT, line 1812: ### Critical: `--cleanup` rewrites a config no live code on this branch writes

This `setup_complete` reset is dead code on `jashan/cleanup-all`. The only producers/consumers of `<persistent>/backend_volume/user_config.json` anywhere on this branch are this block and a single test fixture (`test_run_cleanup.py:112`, which also has the path wrong — see my other comment). The backend that would persist `setup_complete` lives in PR #750 (`jashan/env-fix`, "Move first-run setup and runtime secrets into the UI"), which is still open and not an ancestor of this branch or `dev`. So `os.path.exists(user_config_path)` is always false in practice, and the `try/except PermissionError` + corrupted-JSON branches at L1820–1828 are unreachable.

Worth flagging: the `jashan/env-fix` branch has commit `d8b94a1` (`Revert --cleanup setup_complete reset block`), whose message states the behavior "is being moved to the --cleanup-all PR (#799)". So this block is deliberately landing here ahead of #750, not by accident — please disregard my earlier "two options" framing.

Given that, the one ask is to make this block obviously-defensive rather than silently dead:

1. Add a brief `# NOTE: no-op until #750 lands (#750's revert references this PR)` above L1812 so the next reader doesn't waste time tracing why the block never fires.
2. Add at least one test that creates `user_config.json` at the correct path (`<persistent>/backend_volume/user_config.json`) with `setup_complete: true` + a sibling key, runs `--cleanup`, and asserts the key is stripped while siblings survive (covered in my other comment). Today the block is uncovered, so a regression after #750 lands would ship undetected.

run.py
```diff
@@ -1663,13 +1663,379 @@ def display_welcome_banner():
...
+    print(f"\n{C_BOLD}The following will be PERMANENTLY DELETED:{C_RESET}\n")
+
+    if existing:
+        path_w = max(len(os.path.relpath(p, TT_STUDIO_ROOT)) for _, p, _, _ in existing)
+        path_w = min(max(path_w, 32), 56)
+        for emoji, path, desc, size in existing:
+            rel = os.path.relpath(path, TT_STUDIO_ROOT)
+            size_str = _format_bytes(size) if size > 0 else "—"
+            print(f"  {emoji} {rel:<{path_w}}  {size_str:>10}")
+            print(f"       └─ {C_CYAN}{desc}{C_RESET}")
```
stisiTT, line 1990: ### Critical: `--cleanup-all` leaves SDK-launched containers running

`_cleanup_runtime` only tears down the tt-studio compose stack and then tries `docker network rm tt_studio_network`. But several containers are launched directly via the Docker SDK and attached to `tt_studio_network` outside compose:

- `app/backend/docker_control/docker_utils.py:465` — agent container, `network='tt_studio_network'`
- `app/backend/docker_control/docker_utils.py:176` — `_run_direct_container()` (face-recognition), uses `backend_config.docker_bridge_network_name` → `tt_studio_network`
- Model containers (Llama / Whisper / TTS) launched on the inference-server side land on the same network

`docker compose down -v` doesn't touch any of these, so they survive `--cleanup-all`. The follow-up `docker network rm` then fails (network has active endpoints), and the failure is silently swallowed — leaving a "fresh-clone" state that's actually half-cleaned.

Sketch of what to do before the `network rm` call (adjust to your existing `run_docker_command` helper):

```python
# Stop & remove containers still attached to tt_studio_network (agent,
# face-recognition, and model containers launched via the Docker SDK
# outside compose are not torn down by `docker compose down -v`).
try:
    inspect = subprocess.run(
        ["docker", "network", "inspect", "tt_studio_network",
         "--format", "{{range .Containers}}{{.Name}}\n{{end}}"],
        capture_output=True, text=True, check=False,
    )
    names = [n for n in inspect.stdout.splitlines() if n.strip()]
    for name in names:
        print(f"Stopping container on tt_studio_network: {name}")
        subprocess.run(["docker", "stop", name], check=False)
        subprocess.run(["docker", "rm", "-f", name], check=False)
except FileNotFoundError:
    pass  # docker CLI missing — nothing to clean
```

And surface (not swallow) failures from the subsequent `docker network rm`.

run.py
```diff
@@ -1663,13 +1663,379 @@ def display_welcome_banner():
...
+
+
+def _remove_tt_studio_network_containers(has_docker_access):
+    """Force-remove every container attached to tt_studio_network.
+
+    Deployment containers (vLLM, YOLO, stable-diffusion, …) are spawned outside
+    docker-compose by the backend via docker-control-service, so `compose down`
+    never sees them. They all join `tt_studio_network`, which makes the network
+    a reliable filter. Returns count removed.
+    """
```
stisiTT, line 1837: ### High: `--cleanup-all --yes` deletes `HOST_PERSISTENT_STORAGE_VOLUME` with no path-safety guard

`host_persistent_volume` is read straight from `.env` via `get_env_var` (`run.py:884`) with no validation, and `_remove_path` (`run.py:1703`) has no containment check — it just does `os.path.exists` + sudo-fallback. With `--yes` skipping the confirmation prompt, a typo'd or repurposed `HOST_PERSISTENT_STORAGE_VOLUME` (e.g., `$HOME`, `/data`, `/`, a bind-mount root) gets silently `rm -rf`'d by anyone running `python run.py --cleanup-all --yes` in CI or a script. The interactive prompt is the only thing protecting against this today, and `--yes` removes it.

Suggest gating the delete on a sentinel-or-containment check that runs **even with `--yes`** (the flag should skip the prompt, not the safety check):

```python
def _validate_persistent_volume(path: str) -> str:
    """Refuse to delete anything that isn't clearly a tt-studio volume."""
    resolved = os.path.realpath(path)
    forbidden = {"/", os.path.realpath(os.path.expanduser("~")), "/tmp", "/var", "/home"}
    if resolved in forbidden or resolved == "":
        raise SystemExit(f"Refusing to delete unsafe path: {resolved!r}")
    inside_repo = resolved.startswith(os.path.realpath(TT_STUDIO_ROOT) + os.sep)
    has_sentinel = any(
        os.path.exists(os.path.join(resolved, marker))
        for marker in ("backend_volume", "chroma", "tt_studio_persistent_volume")
    )
    if not (inside_repo or has_sentinel):
        raise SystemExit(
            f"Refusing to delete {resolved!r}: not under TT_STUDIO_ROOT and "
            "no tt-studio sentinel directory found."
        )
    return resolved
```

Then call `host_persistent_volume = _validate_persistent_volume(host_persistent_volume)` immediately after the `get_env_var` line.

test_run_cleanup.py
```diff
@@ -0,0 +1,288 @@
...
+                    run.cleanup_resources(args)
+
+            self.assertEqual(called, {"runtime": False, "images": False, "sentinel": False})
+
+    def test_cleanup_all_yes_removes_state_and_arms_browser_wipe(self):
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            persistent = root / "tt_studio_persistent_volume"
+            persistent.mkdir()
+            (persistent / "user_config.json").write_text('{"hf_token": "hf_test"}')
```
stisiTT, line 112: ### High: Commit `6ccb82f` ships untested; this fixture has the wrong path

Two issues here:

**1. Wrong path.** The production code at `run.py:1812` reads `<persistent>/backend_volume/user_config.json`, but this fixture places the file at `<persistent>/user_config.json` (missing the `backend_volume/` segment). So even if a `--cleanup` test existed, it wouldn't exercise the real code path.

**2. No coverage at all for `6ccb82f`.** `setup_complete` doesn't appear anywhere in `test_run_cleanup.py`. The five existing tests all hit `--cleanup-all`, not the single-dash `--cleanup` path that owns this reset. The `PermissionError` fallback (`run.py:1823`) and corrupted-JSON branch (`run.py:1827`) are also unreachable from any test.

Suggest adding a focused test for the `--cleanup` path:

```python
def test_cleanup_resets_setup_complete_preserves_other_keys(tmp_path, monkeypatch):
    persistent = tmp_path / "tt_studio_persistent_volume"
    cfg_dir = persistent / "backend_volume"
    cfg_dir.mkdir(parents=True)
    cfg_path = cfg_dir / "user_config.json"
    cfg_path.write_text(json.dumps({"setup_complete": True, "hf_token": "hf_test"}))

    monkeypatch.setenv("HOST_PERSISTENT_STORAGE_VOLUME", str(persistent))
    run_cleanup(argparse.Namespace(cleanup=True, cleanup_all=False, yes=True, no_sudo=True))

    assert cfg_path.exists(), "user_config.json should survive --cleanup"
    data = json.loads(cfg_path.read_text())
    assert "setup_complete" not in data
    assert data["hf_token"] == "hf_test"
```

Plus fix the path on this line to include the `backend_volume/` segment so the existing assertion exercises the production code path.

Note: this comment is tightly coupled to my other comment about `6ccb82f` being dead code on this branch — if that commit gets deferred to land alongside #750, this test goes with it.

stisiTT (dismissed):
Requesting changes — see inline comments on the prior review (#pullrequestreview-4340705944).

**Blocker — must fix before merge:**

- **Critical (functional hole)** — `--cleanup-all` leaves SDK-launched containers running (`run.py:1955`). The PR's stated purpose is "reset TT Studio to fresh-clone state," but model/agent/face-recognition containers attached to `tt_studio_network` outside the compose stack survive `docker compose down -v` and hold board/GPU resources. The subsequent `docker network rm tt_studio_network` then fails and the failure is silently swallowed. Needs the teardown loop in the linked comment before `network rm`. The feature doesn't deliver what its name and help text claim until this is fixed.

**Strong recommendation — not blocking, but please consider:**

- **High (safety)** — `--cleanup-all --yes` has no path-safety guard on `HOST_PERSISTENT_STORAGE_VOLUME` (`run.py:1834`). Under the happy path this is fine, but a misconfigured/repurposed env value gets silently `rm -rf`'d under `--yes` in CI or scripted use. The sentinel-or-containment check from the linked comment is small and cheap.

**Non-blocking comments — address in this PR or as follow-up:**

- **Critical (code hygiene, dead-today-by-design)** — `--cleanup` rewrites a `user_config.json` that no live code on this branch writes (`run.py:1812`). Per #750's revert message this block is intentionally landing here ahead of #750. Ask is narrowed to a `# NOTE` + a real test exercising the block at the correct path.
- **High (test quality)** — Commit `6ccb82f` has no test coverage and the existing fixture writes to the wrong path (`test_run_cleanup.py:112` uses `<persistent>/user_config.json` while production reads `<persistent>/backend_volume/user_config.json`). Fix the path and add the focused test in the linked comment.

Happy to re-review once the blocker is addressed.

jashansinghTT:
Follow-up `95499f1` addresses two orphaned anonymous volumes left after `--cleanup-all` on the quietbox:

```
$ docker volume ls
local     476754b33a3e55f4aac4c732e63417c22abeee05a31c5d8fa8b60b426bf3002e
local     ade30e1508c3283fdef5b69f507bca17fd2edee8e07fe1ab127eb9e91440806a
```

**Root cause:** content under the mountpoints is unambiguously a node_modules tree (`@alloc`, `acorn`, `aria-hidden`, …) — the frontend dev container's anonymous volume from `app/docker-compose.dev-mode.yml:44` (`- /app/node_modules`). The leak path was: my `_remove_tt_studio_network_containers` runs BEFORE `compose down -v` and calls `docker rm -f` (no `-v`). The frontend dev container is on `tt_studio_network`, so the helper grabs it first and removes the container without its anon volume. By the time `compose down -v` runs, the container is already gone and compose has nothing left to clean up — the anon volume orphans.

**Changes**
- `docker rm -f` → `docker rm -fv` in `_remove_tt_studio_network_containers` so anonymous volumes attached to containers we own are reclaimed inline.
- New `_prune_anonymous_volumes` helper: snapshots `docker volume ls -q` before/after `docker volume prune --force` (no `--all`) and returns the delta. `prune` without `--all` only touches anonymous unused volumes, so named volumes from other projects on the same host are safe. Defensive net for orphans that were created before this fix (e.g. the two on QB now) AND for any future code path that misses `-v`.
- Wired into `cleanup_resources` between model-volume removal and image removal. Only runs on `--cleanup-all`, not plain `--cleanup`.
- Pre-confirmation inventory now lists dangling anonymous volumes.

**Tests**
- Assert `docker rm -fv` (not `-f`) in the network-containers helper test.
- `test_prune_anonymous_volumes_counts_delta`: verifies the helper uses `--force` WITHOUT `--all`/`-a` so named volumes survive, and computes count from the before/after `volume ls` diff.
- `test_prune_anonymous_volumes_zero_when_nothing_changes`: empty-diff path returns 0.

**Live verification on `sjc2-qb2-9b22`** against the two existing orphans:

```
=== BEFORE ===
local     476754b33a3e…
local     ade30e1508c3…

=== Calling _prune_anonymous_volumes ===
removed: 2

=== AFTER ===
(empty)
```

All 13 tests pass.

anirudTT:
> Requesting changes — see inline comments on the prior review (#pullrequestreview-4340705944).
> 
> **Blocker — must fix before merge:**
> 
> * **Critical (functional hole)** — `--cleanup-all` leaves SDK-launched containers running (`run.py:1955`). The PR's stated purpose is "reset TT Studio to fresh-clone state," but model/agent/face-recognition containers attached to `tt_studio_network` outside the compose stack survive `docker compose down -v` and hold board/GPU resources. The subsequent `docker network rm tt_studio_network` then fails and the failure is silently swallowed. Needs the teardown loop in the linked comment before `network rm`. The feature doesn't deliver what its name and help text claim until this is fixed.
> 
> **Strong recommendation — not blocking, but please consider:**
> 
> * **High (safety)** — `--cleanup-all --yes` has no path-safety guard on `HOST_PERSISTENT_STORAGE_VOLUME` (`run.py:1834`). Under the happy path this is fine, but a misconfigured/repurposed env value gets silently `rm -rf`'d under `--yes` in CI or scripted use. The sentinel-or-containment check from the linked comment is small and cheap.
> 
> **Non-blocking comments — address in this PR or as follow-up:**
> 
> * **Critical (code hygiene, dead-today-by-design)** — `--cleanup` rewrites a `user_config.json` that no live code on this branch writes (`run.py:1812`). Per #750's revert message this block is intentionally landing here ahead of #750. Ask is narrowed to a `# NOTE` + a real test exercising the block at the correct path.
> * **High (test quality)** — Commit `6ccb82f` has no test coverage and the existing fixture writes to the wrong path (`test_run_cleanup.py:112` uses `<persistent>/user_config.json` while production reads `<persistent>/backend_volume/user_config.json`). Fix the path and add the focused test in the linked comment.
> 
> Happy to re-review once the blocker is addressed.

I don't see how containers launched outside of TT Studio via the Docker SDK are our responsibility to clean up — `--cleanup-all` should only be accountable for what TT Studio itself started. Can you clarify why you consider this a blocker?

anirudTT:
if we are using AI to write comments back / review them, can we make them more concise, its bit long and hard to follow.

run.py
jashansinghTT: I think, this functionality makes it future proof and doesn't disturb the current flow. I will skip this for now and focus on the other issues.

run.py
jashansinghTT: Thanks for this comment, this was resolved and was flagged by @rnabeelTT as well

jashansinghTT: Basically, this was addressed in a previous commit before this comment, so currently it gets rid of all the volumes at the present.

run.py
jashansinghTT: I think we can scope out this PR later. Currently, we are not saving anything on the servers. It's all on a local machine. Hence, we can have a security check in the future, but for now this simplistic approach is viable. We can scope this out in a later PR, though, and it's a valid suggestion.

anirudTT (changes requested):
Re tested and ran into these issues, pasted screenshots below.

- few logs don't get deleted without sudo access , might not be in scope for this PR / issue and we need to root cause sudo usage
- it did not delete docker volumes for me.

Screenshot 2026-05-21 at 5 14 41 PM
Screenshot 2026-05-21 at 5 15 50 PM

anirudTT:
> Follow-up `95499f1` addresses two orphaned anonymous volumes left after `--cleanup-all` on the quietbox:
> 
> ```
> $ docker volume ls
> local     476754b33a3e55f4aac4c732e63417c22abeee05a31c5d8fa8b60b426bf3002e
> local     ade30e1508c3283fdef5b69f507bca17fd2edee8e07fe1ab127eb9e91440806a
> ```
> 
> **Root cause:** content under the mountpoints is unambiguously a node_modules tree (`@alloc`, `acorn`, `aria-hidden`, …) — the frontend dev container's anonymous volume from `app/docker-compose.dev-mode.yml:44` (`- /app/node_modules`). The leak path was: my `_remove_tt_studio_network_containers` runs BEFORE `compose down -v` and calls `docker rm -f` (no `-v`). The frontend dev container is on `tt_studio_network`, so the helper grabs it first and removes the container without its anon volume. By the time `compose down -v` runs, the container is already gone and compose has nothing left to clean up — the anon volume orphans.
> 
> **Changes**
> 
> * `docker rm -f` → `docker rm -fv` in `_remove_tt_studio_network_containers` so anonymous volumes attached to containers we own are reclaimed inline.
> * New `_prune_anonymous_volumes` helper: snapshots `docker volume ls -q` before/after `docker volume prune --force` (no `--all`) and returns the delta. `prune` without `--all` only touches anonymous unused volumes, so named volumes from other projects on the same host are safe. Defensive net for orphans that were created before this fix (e.g. the two on QB now) AND for any future code path that misses `-v`.
> * Wired into `cleanup_resources` between model-volume removal and image removal. Only runs on `--cleanup-all`, not plain `--cleanup`.
> * Pre-confirmation inventory now lists dangling anonymous volumes.
> 
> **Tests**
> 
> * Assert `docker rm -fv` (not `-f`) in the network-containers helper test.
> * `test_prune_anonymous_volumes_counts_delta`: verifies the helper uses `--force` WITHOUT `--all`/`-a` so named volumes survive, and computes count from the before/after `volume ls` diff.
> * `test_prune_anonymous_volumes_zero_when_nothing_changes`: empty-diff path returns 0.
> 
> **Live verification on `sjc2-qb2-9b22`** against the two existing orphans:
> 
> ```
> === BEFORE ===
> local     476754b33a3e…
> local     ade30e1508c3…
> 
> === Calling _prune_anonymous_volumes ===
> removed: 2
> 
> === AFTER ===
> (empty)
> ```
> 
> All 13 tests pass.

DO we actually need this service to delete volumes or containers outside of what we generated ?

test_run_cleanup.py
jashansinghTT: Acknowledged —> d8b94a1 on #750 already pairs with this. Will add the NOTE comment + test in #750 where it'll be exercisable.

run.py
jashansinghTT: Also the containers addressed here are stopped

stisiTT:
> I don't see how containers launched outside of TT Studio via the Docker SDK are our responsibility to clean up — `--cleanup-all` should only be accountable for what TT Studio itself started.

I think this conflates "outside docker-compose" with "outside TT Studio." Those deployment containers are spawned by the TT Studio backend via `docker-control-service`, join `tt_studio_network`, and their lifecycle records live in `host_persistent_volume`. They only exist because someone deployed via TT Studio. Mechanically, `docker network rm tt_studio_network` also fails until they're cleaned up, so the cleanup leaves a stale network behind otherwise.

Regardless, `4774dd2` already addresses this. It adds `_remove_tt_studio_network_containers` to wipe them before `compose down -v`, and the commit message itself notes they *"are spawned outside docker-compose by the backend via docker-control-service, so `compose down` never saw them."* Considering the blocker resolved; will dismiss the changes-request unless you spot something I'm missing.

stisiTT (approved):
C2 is resolved by `4774dd2` (new `_remove_tt_studio_network_containers` plus `CleanupDockerSurfaceTests`). Marking that inline thread resolved and approving. Three non-blocking follow-ups below.

### C1 (run.py:1918, dead `setup_complete` block)

This block runs today, finds nothing, does nothing. The file it operates on is only written by code in PR #750, which has not merged yet. It is not broken, just dormant until that PR lands.

Ask (non-blocking): add a one-line marker above the block so the next reader does not spend time tracing why it never fires. Suggested wording: `# NOTE: no-op until #750 lands (#750's revert references this PR)`.

### C3 (run.py:1946, no path-safety guard under `--yes`)

`host_persistent_volume` is read raw from `.env` and passed straight into `_remove_path`. With `--yes` skipping the confirmation prompt, a misconfigured or repurposed `HOST_PERSISTENT_STORAGE_VOLUME` gets silently `rm -rf`'d in CI or scripted use. The prompt is currently the only safeguard against a typo or misuse.

Ask (strong recommendation, still non-blocking): add a guardrail before the destructive loop. Refuse to delete unless the path either lives inside `TT_STUDIO_ROOT` or contains a tt-studio sentinel directory (`backend_volume`, `chroma`, or `tt_studio_persistent_volume`). Should run even with `--yes`.

### C4 (test_run_cleanup.py:112, wrong fixture path plus no `setup_complete` coverage)

The fixture writes `user_config.json` at `<persistent>/user_config.json`, but production code reads it from `<persistent>/backend_volume/user_config.json`, so the production code never sees the fixture. Separately, the C1 reset block has zero coverage in the test file (grep for `setup_complete` returns 0 hits).

Ask (non-blocking): two small changes. Fix the fixture path so it includes `backend_volume/`. Add one focused test that creates the file with both `setup_complete` and a sibling key, runs the `--cleanup` (not `--cleanup-all`) path, and asserts `setup_complete` is gone while the sibling key is preserved.

anirudTT (approved):
lgtm . Tested it again after the updates and it works well . Great work!!!

PR #805, opened 2026-05-20
chore: depreciate startup script

anirudTT:


rnabeelTT (approved):
Works just fine! no major changes

PR #812, opened 2026-05-25
Deployment progress bar for Media models

rnabeelTT:
## Description

Builds on the LLM deploy-progress work already in dev (#758) to add progress reporting for tt-media-inference-server models (Whisper, SpeechT5, etc.). 

> **Note:** The LLM path remains completely unchanged.

---

## Updated Look on Models Deployed Page
Screenshot 2026-05-27 at 1 17 56 PM

### Media-Specific Phase Classifier 
`app/backend/model_control/log_classifier.py`
* **Trimmed 6-Phase Template:** Verified against real `distil-large-v3` / `speecht5_tts` logs:
* **Early Bookkeeping Collapse:** Settings init, Prometheus setup, and worker spawns are collapsed into `container_starting` since they all fire before the frontend mounts.
* **HuggingFace Detection:** Adds `_DOWNLOAD_MEDIA_HF_LOAD_RE` to pick up the `Loading HuggingFace model: <repo>` line emitted by `transformers.from_pretrained` (Whisper/SpeechT5 do not go through the standard wrapper that emits `Downloading weights for model:`).
* **Sub-step Interpolation:** Sub-step counters (`MEDIA_LOAD_TOTAL`, `MEDIA_WARMUP_TOTAL`) interpolate progress through the two longest phases. This ensures the bar advances during otherwise-silent model loads and the multi-size encoder warmup loop.

### Download-Progress Helper 
`app/backend/model_control/download_progress.py`
* **Cache Path Derivation:** `_media_container_path()` derives the HF hub cache path from a repo ID, as media logs only surface the repo rather than a target path.
* **95% Threshold Heuristic:** Flags weights as cached when `du -sb` shows the cache is effectively full, catching edge cases where no explicit "cached" log line was fired.

### Monotonic Progress Floor 
`app/backend/model_control/views.py`
* **Phase Latching:** `_apply_phase_latch` now also latches `max_progress`. This prevents the progress bar from visually regressing when an earlier substep marker scrolls out of the 200-line log tail.
* **Empty Log Tail Support:** Empty log tails still run the classifier (defaulting to `phases[0]`) so the latched maximum can re-emit, instead of returning `None` mid-warmup.

### Log Stream Noise Filter 
`docker-control-service/routers/containers.py`
* **Access Log Suppression:** Drops `uvicorn INFO: <ip>:<port> - "GET /health HTTP/1.1" 200`-style access lines from the UI Logs tab stream (`follow=True` only). 
* *Note: One-shot tails used by the backend classifier remain untouched.*

---

## Test Plan

- [ ] **Cold Whisper Deployment:** Deploy a Whisper model (`distil-large-v3`) cold — confirm the progress bar walks through all 6 media phases sequentially without regressing.
- [ ] **SpeechT5 Interpolation:** Deploy a SpeechT5 model — confirm `loading_weights` and `warming_up` advance smoothly via sub-step interpolation.
- [ ] **Cached Weights:** Deploy with weights already cached — confirm the 95% heuristic flags them as cached and the bar briefly pins to 100%.
- [ ] **Regression Testing:** Deploy any LLM — confirm the LLM path is completely unaffected.


anirudTT (approved):
good to go from my end , @jashansinghTT can you test the voice agent ?

PR #814, opened 2026-05-26
Consolidate the state seen by reads to Single Source of Truth

### LINKED ISSUE #787: Tracking: Reset board flow performance + deployed-model state consolidation
## Summary

Tracking issue for two linked pieces of work around the Reset Board flow and deployed-model state. Filed together because they share a root cause: there is no single source of truth for what's deployed on the board, which makes the reset flow both slow and unreliable.

## Sub-issues

- #779 — **Bug:** Reset board flow is slow (sequential model stops) and progress UI can desync from real state. User-visible symptom; can be partially fixed in isolation (parallelize stops, persistent reset lock).
- #786 — **Feature/refactor:** Consolidate deployed-model state — make `deployment_store` the single source of truth. Removes the structural reason #779 exists. Required to fully eliminate "modal lists ghost models" and similar drift symptoms.

## Suggested order

1. #779 first for the user-visible wins that don't depend on consolidation (parallel `Promise.all` on stops, persistent reset-in-flight lock, better progress copy).
2. #786 second to remove the underlying drift and let #779's remaining gaps (ghost models in the reset modal) close naturally.

## Context

Surfaced from a user observation that the "Resetting all chips…" modal takes longer than it should and that there appears to be no single place that records "what is deployed." Investigation confirmed:

- Reset flow stops models sequentially and re-derives the deployed list from live Docker on every click.
- A persistent registry (`deployment_store.py` → `deployments.json`) exists but is not the source of truth — it competes with a `deploy_cache` LocMemCache, live Docker listings, and a frontend `ModelsContext` merge.
- Two independent writers update the registry; unexpected-death detection lags by up to 60s ([health_monitor.py:144](app/backend/docker_control/health_monitor.py#L144)).

Recent commit `4cf4332` introduced unexpected-death surfacing via `health_monitor`; this tracking issue is the natural follow-up to make that real-time and propagate the same source of truth through the reset UI.

rnabeelTT:
The question of which models are currently deployed was answered by four uncoordinated sources (deployment_store, deploy_cache LocMemCache, live Docker, frontend ModelsContext merge) that routinely disagreed — potentially producing ghost models in the reset modal, a brittle partial-name join in the frontend, and a 60s death-detection lag. This PR routes every read through a single backend function and migrates every UI consumer to it.

**Backend** 
- get_canonical_deployments() in docker_utils.py is the new SoT. It joins deployment_store records with live Docker by matching container_id OR container_name (the name fallback is load-bearing for CHAT-model job_id placeholder windows during multi-hour weights downloads), then enriches with model_impl from the catalog. It reconciles ghost records (running-but-no-container, beyond grace) to status="stopped" as a read-time side-effect, so stale rows die at the source.
- New endpoint GET /docker-api/deployments/ exposes the canonical view via DeploymentsView.
/docker-api/status/, /models-api/deployed/, /docker-api/chip-status/ are now thin shims over the canonical function. Response shapes preserved byte-for-byte for backwards compatibility.
- update_deploy_cache() simplified to a thin memoization of canonical output. Existing callers of get_deploy_cache() keep their Python-object model_impl semantics — required by _resolve_model_identity and downstream inference views.
- ChipSlotAllocator._get_active_deployments now delegates to canonical instead of its own reconciliation loop. Reset modal, deployed-models table, and allocator now see the same world.
- Health monitor poll: 60s → 5s. Read-time reconciliation in canonical now catches death on next status fetch; this thread becomes the backstop, not the sole path.

**Frontend**
- fetchDeployments() + CanonicalDeployment type added.
- ModelsContext.refreshModels rewritten: single fetch, no merge, exact container_id join. Eliminates the partial-name-substring match ("Llama" previously matched both "Llama-3.1-8B" and "Llama-3.1-70B") and the orphan-row case where /models-api/deployed/ returned a record without a Docker match.


anirudTT (approved):
lgtm

PR #829, opened 2026-05-30
fix: make reset card theme-aware

AnkittChauhan:
## Summary
Fixes the Reset Card dialog so it follows the app theme in both light and dark mode.

The reset dialog previously used hard-coded dark Tailwind classes, which caused the card, step rows, status banners, warning panel, command output, and footer buttons to remain dark even when TT-Studio was switched to light mode.

## Changes
- Updated `ResetIcon.tsx` reset dialog shell to use light defaults with `dark:` variants.
- Made reset step rows theme-aware for pending, active, done, and skipped states.
- Updated board status banners so healthy, missing-device, resetting, and bad-state messages are readable in light mode.
- Updated warning, failure, command output, close, and cancel UI styles for both themes.
- Preserved the destructive reset button color semantics.

## Validation
- Verified locally in dark mode that the Reset Card still matches the existing dark UI.
- Verified locally in light mode that dialog background, text, rows, borders, panels, and buttons switch to light styling.
- Ran `npm run type-check` from `app/frontend`.

## Working

https://github.com/user-attachments/assets/47d27b3b-1c79-421b-b4b2-5c775aaf4ca5

### DIFF
--- a/app/frontend/src/components/ResetIcon.tsx
+++ b/app/frontend/src/components/ResetIcon.tsx
@@ -68,23 +68,23 @@ function StepRow({
   return (
     <div
       className={`flex items-start gap-3 p-3 rounded-lg border transition-all duration-300 ${state === "active"
-          ? "bg-blue-900/30 border-blue-500/40"
-          : state === "done"
-            ? "bg-green-900/20 border-green-600/30"
-            : state === "skipped"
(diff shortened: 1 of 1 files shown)


jashansinghTT (commented):
Works good on the UI

anirudTT (approved):

Issue #825, opened 2026-05-29
LLM deployments get torn down by --dev / --cleanup, but speech models survive


jashansinghTT:
## What happens

After deploying a mix of models (e.g. Whisper STT, a Distil/TTS model, and an LLM), if I stop the stack with `--cleanup` or restart into dev mode with `--dev`, the **LLM disappears** — it shows as dead/stopped in the UI. The **speech models (TTS / STT) stay deployed and usable**.

## Steps to reproduce

1. Deploy a Whisper (STT) model, a TTS model, and an LLM.
2. Run `python3 run.py --cleanup` (or restart into dev with `python3 run.py --dev`).
3. Bring the UI back up.

**Result:** the LLM is gone; the speech models are still there.
**Expected:** deployed models should be treated the same — the LLM shouldn't be singled out and killed.

## Likely cause

LLM (CHAT) models are deployed **through the TT Inference Server FastAPI** (`start_chat_deployment` → `/run`), so their lifecycle is tied to that server. Both `--cleanup` and the `--dev` restart kill/restart that FastAPI server (`cleanup_fastapi_server`), which takes its LLM containers down with it. Speech models are run as standalone containers, so they're unaffected.

## Files that might be responsible

- `run.py` — `cleanup_resources()` (~L1666): `docker compose down -v`, `docker network rm tt_studio_network` (~L1687), then `cleanup_fastapi_server()` (~L1696)
- `run.py` — `cleanup_fastapi_server()` (~L3337) kills the TT Inference Server process on :8001; `start_fastapi_server()` (~L3165) is restarted on `--dev` (~L4632)
- `app/backend/docker_control/tt_inference_client.py` — `start_chat_deployment()` (~L24): LLMs deployed via the inference server
- `app/backend/docker_control/docker_utils.py` — model-type branching (~L296 / L322 / L357) where CHAT vs TTS/SPEECH paths diverge
- `app/backend/docker_control/health_monitor.py` — `check_container_health()` (~L76) marks the LLM as dead/exited afterward
- `app/backend/docker_control/deployment_sync.py` / `deployment_store.py` — deployment state tracking

PR #833, opened 2026-06-02
fix: preserve deployment containers across --cleanup; --cleanup-all still removes them

jashansinghTT:
## Summary

Deployment containers (LLM via vLLM, TTS, STT) were being torn down on every `--cleanup`, even though the host-side teardown (compose containers, FastAPI, Docker Control Service) doesn't need them gone. The Tenstorrent device-resident state takes minutes to re-load per LLM, so this turned routine restarts into expensive re-initializations.

After this change:
- **`--cleanup`** — tears down host services + compose containers + clears the welcome flag, but **deployment containers stay running** and stay attached to `tt_studio_network`. When TT Studio comes back up, the backend re-discovers them via the existing canonical-reconcile path.
- **`--cleanup-all`** — unchanged: still tears down deployments, the network, host services, persistent volumes, etc. (full reset).

### Changes
- **`run.py`** — `_cleanup_runtime` now branches on `args.cleanup_all`. When False, it skips `_remove_tt_studio_network_containers` and the explicit `docker network rm tt_studio_network`. When True, behaviour is identical to before.
- **`test_run_cleanup.py`** — split the previous deployment-ordering test into two cases (basic vs. full cleanup), and refreshed the two pre-existing tests whose fixtures had drifted from the image-ancestor union added in #825.

### How this addresses the original parity goal
The earlier direction was "both LLM and voice get cleaned up reliably on `--cleanup`". This PR inverts that: both LLM and voice now **survive** `--cleanup`, and both are still removed by `--cleanup-all`. Survival is the property that matters in practice — restarts shouldn't cost device re-load time. The full-reset path is preserved for when the user actually wants a fresh state.

## Test plan
- [x] `python -m unittest test_run_cleanup` — all 14 pass
- [x] Real-docker e2e: dummy container on `tt_studio_network` survives `_cleanup_runtime(cleanup_all=False)` and is removed by `_cleanup_runtime(cleanup_all=True)`
- [x] Deploy an LLM model, run `python run.py --cleanup`, then `python run.py` again — LLM container should remain running and reachable
- [x] Deploy a TTS/STT model, same flow — voice container should remain running
- [x] Deploy both, run `python run.py --cleanup-all` — both removed along with everything else


anirudTT (approved):
LGTM. Tested voice agent deploy → `python run.py --cleanup`; cleanup left all model containers running, including LLMs.

PR #835, opened 2026-06-02
Add WAN T2V, Flux, Voice Agent, and inference-api reliability fixes

### LINKED ISSUE #857: flux + wan models working in tt-studio

stisiTT:
Brings Wan T2V and Flux video generation to tt-studio, plus two runtime fixes discovered during testing.

---

## Video generation (Wan T2V / Flux)

- Backend: async submit → poll → download flow so the request doesn't block while the video renders
- Frontend: video gen page, chat component, input area, progress bar, hooks, nav/route wiring
- Progress estimate raised to 4m15s to better match actual Wan/Flux render times

---

## Fix: multi-chip device deployment

Whole-board multi-chip media models (p300x2, p150x4, p150x8, t3k, galaxy) were deployed with a single `device_id="0"`, mounting only one chip while the mesh required all of them. `run_container` now omits `device_id` when `chips_required > 1`, matching the existing CHAT path. Verified on P300x2: container now mounts the full board and device init reports all 4 chips.

---

## Fix: catalog sync (sync_models_from_inference_server.py)

P300/P300x2 entries were silently dropped from the synced catalog due to a missing device-type mapping and casing mismatch. Adds the mappings with `.upper()` normalisation, coerces env var values to `str` for Docker compatibility, and uses a stdlib version comparator to pick the highest-version entry per model.

---

## Fix: Voice Agent Llama deploy and LLM streaming

Two bugs prevented Llama-3.1-8B-Instruct from working in the Voice Agent pipeline:

**Deploy failure — missing `--host-hf-cache`**
`hf_cache_models.json` only allowlisted media models. Without `--host-hf-cache` the weights symlink map can't resolve and vLLM aborts with "Unrecognized model". Added `Llama-3.1-8B-Instruct` to the allowlist.

**Runtime failure — async/sync mismatch in pipeline_views.py**
`pipeline_views.py` iterated the async generator `stream_response_from_external_api` with a sync `for`-loop, raising `TypeError: 'async_generator' object is not iterable`. The error was silently swallowed so the LLM step always failed. Replaced with a synchronous SSE-parsing `requests.post` stream that carries the bearer JWT and extracts `choices[].delta.content`, matching the STT/TTS steps in the same view.

---

## Fix: inference-api restart drops MODEL_SPECS_ENV (stopgap)

The inference-api reads `MODEL_SPECS_ENV` once at startup to select the `dev` or `prod` model-spec YAML catalog. The dev `video.yaml` declares `TT_DIT_CACHE_DIR` for Wan media models; prod does not.

When the process was restarted (to reload the Llama allowlist), it defaulted to `prod`, Wan deployed without the cache dir, and the transformer weight load ran uncached and hung indefinitely.

Pins `MODEL_SPECS_ENV=dev` in `start_fastapi_server()` so the catalog choice is deterministic across restarts, while still honoring an explicit operator override (`MODEL_SPECS_ENV=prod` takes precedence).

> **Stopgap:** remove this block once `TT_DIT_CACHE_DIR` lands in the prod catalog upstream in tt-inference-server.

---

## Fix: guard against partial HF weight cache causing Wan/media container crash

A prior interrupted download (network drop, XET 403 loop) can leave an HF cache snapshot in a split state: some shards fully downloaded (symlinks in the snapshot dir) while earlier shards are still `.incomplete` blobs. On re-deploy, `check_model_weights_dir` found the existing safetensors symlinks, declared setup complete, skipped the download, and launched the container — which immediately crashed with `FileNotFoundError` for the missing shards.

Two layers of defense added to `inference-api/api.py`:

- **Incomplete-blob guard:** monkey-patches `HostSetupManager.check_model_weights_dir` to check for `*.incomplete` files in the HF cache `blobs/` directory (two levels above the snapshot dir) before declaring a snapshot ready. If any are found, returns `False` to force `hf download` to run and resume the partial blobs.
- **Synchronous download:** adds `HF_HUB_DISABLE_XET=1` to the `setup_host` environment so `hf download` uses plain HTTPS rather than XET's async blob scheduling. XET can return exit 0 before background blob transfers finish; HTTPS does not. The container already receives this flag via the catalog patch; this extends it to the host-side setup download.


anirudTT (commented):
https://github.com/tenstorrent/tt-studio/issues/958
seeing the issues posted to the ticket here , @stisiTT could you help me out?

anirudTT:
This is the issue i get testing this now, do you have a different branch that you want us to use?
```
> ✅ Successfully removed invalid artifact directory with sudo
> 📥 Downloading TT Inference Server from GitHub branch: stisi/fix-persistent-tt-dit-cache-dir
>    Downloading from: https://github.com/tenstorrent/tt-inference-server/archive/refs/heads/stisi/fix-persistent-tt-dit-cache-dir.tar.gz
>    This may take a few minutes...
> ⛔ Branch 'stisi/fix-persistent-tt-dit-cache-dir' not found on GitHub (HTTP 404).
>    The branch name you configured does not exist.
>    You entered: TT_INFERENCE_ARTIFACT_BRANCH=stisi/fix-persistent-tt-dit-cache-dir
>    Run: python run.py --reconfigure-inference-server
>    Valid branches: https://github.com/tenstorrent/tt-inference-server/branches
> ⛔ Cannot start TT Studio: TT Inference Server setup failed. Exiting.
```

stisiTT:
Hey @anirudTT — the branch was deleted but the fix commit still exists. I've updated the code to support pinning by commit SHA directly, so it'll now download the archive by SHA instead of branch name. Just pushed — should work now.

app/backend/docker_control/docker_utils.py
```diff
@@ -118,6 +118,11 @@ def _ensure_network():
     "unknown": "cpu",
 }
 
+# Inference-server device names that denote a single chip/card (as opposed to a
```
anirudTT, line 122: can we decouple this from this PR , to help merge it faster?
I think these are not bugs we see anymore or they need to ideally come in via a new PR.

app/backend/docker_control/docker_utils.py
```diff
@@ -296,6 +301,14 @@ def infer_inference_server_device(impl, board_type=None):
...
     if chips_required == 1:
         device = _BOARD_TO_SINGLE_CHIP_DEVICE.get(board_type, "cpu")
+        # A constituent single-chip device (e.g. p150 on a P300x2 board) is only valid
+        # if the model actually declares support for it. Media models like FLUX have no
+        # p150 spec — only the whole-board mesh (p300x2). When the single chip isn't a
+        # supported device for this model but the whole board is, deploy on the board mesh.
+        supported = {map_board_type_to_device_name(cfg.name) for cfg in impl.device_configurations}
+        board_device = map_board_type_to_device_name(board_type)
+        if device not in supported and board_device in supported:
+            device = board_device
```
anirudTT, line 312: same here ,

> can we decouple this from this PR , to help merge it faster?
> I think these are not bugs we see anymore or they need to ideally come in via a new PR.

app/backend/docker_control/docker_utils.py
```diff
@@ -349,9 +362,19 @@ def run_container(impl, weights_id, device_id=0, host_port=None, use_image_overr
         payload["service_port"] = str(BASE_SERVICE_PORT + primary_device_id)
         service_port = BASE_SERVICE_PORT + primary_device_id
 
-        # Pin to specific chip slot(s). For multi-chip single-card mode this is a
-        # comma-separated string that the inference server passes as --device-id 0,1.
-        payload["device_id"] = str(device_id)
+        # Pin to a specific chip slot only for single-chip models. For multi-chip
```
anirudTT, line 366: this as well , i suspect this is solved when you re-base to dev

app/backend/shared_config/sync_models_from_inference_server.py
```diff
@@ -173,6 +193,41 @@ def pick_higher_status(current: str | None, candidate: str) -> str:
     return current if STATUS_ORDER.get(current, 0) >= STATUS_ORDER.get(candidate, 0) else candidate
 
 
+def _version_key(entry: dict) -> tuple[int, tuple[int, ...]]:
+    """Sort key for selecting the highest-version device entry.
```
anirudTT, line 197: is this a fix for the run.py to parse the right inference server version?

anirudTT (changes requested):

app/backend/docker_control/docker_utils.py
stisiTT: This one isn't on `dev` in any form, so it won't come in via a rebase — it's net-new here. It's also load-bearing for the multi-chip media (WAN/Flux) deploys this PR adds: the single-chip device set is what gates the `device_id` behavior below. I can split it into a follow-up PR, but it's bundled in `8794e78` alongside the WAN/Flux `models.json` device entries, so decoupling means a hunk-level split rather than dropping a commit. Happy to do that if you'd prefer a smaller PR — just flagging it's not free, and it's not a stale fix.

app/backend/docker_control/docker_utils.py
stisiTT: Same situation as above — `board_device in supported` doesn't exist on `dev`, so it's not superseded by a rebase. It's part of the same commit (`8794e78`) as the single-chip device handling. I'll bundle the decouple decision for both of these together.

app/backend/docker_control/docker_utils.py
stisiTT: Checked this against `origin/dev` — it's actually *not* solved by a rebase. `dev` still sets `payload["device_id"] = str(device_id)` unconditionally; this change gates it behind single-chip devices so multi-chip media deploys don't get a bogus `device_id`. So if anything, taking dev's version on rebase would silently revert the fix. This one is cleanly isolated in `c457077` (single-file commit), so it's the easiest to lift into a separate PR if we go that route.

app/backend/shared_config/sync_models_from_inference_server.py
stisiTT: Not quite — this isn't the run.py inference-server-version parser. It's a per-device "pick the canonical entry" helper that selects the highest-semver entry (and its `version`/`docker_image`) when the catalog has multiple entries for the same model/device. Stdlib-only, no run.py coupling. It's independent of `dev` and auto-merges cleanly with #972's catalog trimming.

stisiTT:
Thanks for the review @anirudTT. I ran the divergence against `dev` to check the "solved on rebase" theory: the branch is 17 ahead / 20 behind, and a merge only produces 2 trivial keep-both conflicts (NavBar / route-config — our video-gen route vs the new coding-agents route).

On the `docker_utils.py` items specifically — none of the three are on `dev`, so a rebase won't bring them in or supersede them; the device_id one (L365) would actually be *reverted* by taking dev's side. They're decoupleable into a follow-up, but only `c457077` (L365) lifts out as a clean commit — the other two are interleaved with the WAN/Flux `models.json` work, so it's a hunk-split. My preference is to keep them since they're load-bearing for the multi-chip media deploys this PR enables, but I'm happy to split if a smaller PR is the priority — your call.

anirudTT:
Logged the deploy / reliability problems surfaced in review and testing as tracking issues:

- #997 — **Flux won't deploy correctly on mesh boards (P300x2)** — device routing fixed here (`8794e78`, `c457077`); the I2V base-URL 404 is split out below.
- #998 — **Wan2.2-T2V deploys unreliably** — uncached hang / false "Died Unexpectedly" stopgap-fixed here; log streaming + upstream artifact-pin durability still open.
- #999 — **HF media-weights download is fragile** — host-side guards fixed here (`94438b9`); the `apply_media_catalog_env_overlay()` call to a missing `patch_catalog_env.py` is still dead code.
- #1000 — **I2V models don't work (status/download 404)** — open; blocks all I2V deploys.

Open items that are effectively blockers for this PR: the I2V base-URL 404 (#1000) and the missing `patch_catalog_env.py` referenced by run.py (#999).

anirudTT (approved):
approving, we will need to address some of these issues posted above before we open up our rc.

Issue #836, opened 2026-06-03
Qwen shows incorrect Device ID


rnabeelTT:
### The Bug
Qwen3-32B is deployed across all 4 devices. In the Models Deployed page, it should therefore show all 4 device ID's, which in the normal case using the QB2 are 00, 01, 02, 03. Currently, it shows only 00.

rnabeelTT:
fixed by PR #837

PR #837, opened 2026-06-03
Fix to show correct Qwen3-32B Device IDs on the Models Deployed Page

rnabeelTT:
### Summary
The Models Deployed page showed the wrong device IDs for multi-chip models other than Llama - only the primary/base slot was recorded instead of the full set of slots the model actually occupies.

This PR fixes that by computing the complete set of chip slots a deployment uses and persisting it on the ModelDeployment record:
- Multi-chip models occupy chips_required contiguous slots starting at the allocated base slot (clamped to the board size).
- Forced full-board Llama occupies every slot on the board.
- Single-chip / explicit multi-slot requests keep the exact allocated slot list.

### Changes
- [docker_utils.py]: derive the full slot range from chips_required when recording the deployment.
- [views.py]: compute occupied_device_ids and store it on the deployment (instead of the request-only device_ids); also log it.

### Testing
Deploy Qwen3-32B and check the device IDs shown on its row in the Models Deployed page — they should now reflect all the slots the model occupies.

### Updated View
image

**Fixes Issue #836**


anirudTT (approved):
lgtm

PR #841, opened 2026-06-03
Quicker parallel delete and reset for individual multi-slot model deployments

rnabeelTT:
## Speed up per-model delete for multi-slot deployments
Changes the device-reset step of the per-model delete flow from a per-chip sequential loop to a single batched tt-smi -r <id1> <id2> ... call across all of the model's chips at once.

Previously, deleting a model that spanned multiple chips ran tt-smi -r once per chip, one after another. The batched call lets tt-smi coordinate the PCIe secondary-bus reset and ethernet bring-up across all chips in one pass — faster, and safer since it avoids disturbing neighbouring chips' links on every invocation.

The reset's line-read timeout now scales with board size (max(90, 30 × num_devices)), and success/failure reporting is simplified to a single batched outcome.

### File changed: 
[app/backend/docker_control/views.py]: StopStreamView replaced the per-chip loop that ran tt-smi -r once per device with a single batched tt-smi -r <id1> <id2> ... invocation across all of the model's chips, and bumped the line-read timeout to scale with board size (max(90, 30 × num_devices)). Success/failure tracking was collapsed from per-device lists into a single batched outcome.

### Test plan
Deploy any model that takes up multiple slots, then delete it from the Models Deployed page using the Delete button. You should notice a speedup in the device-reset step.


anirudTT (approved):
lgtm

PR #847, opened 2026-06-03
hot-fix: persist user preference for terms acceptance during first run

anirudTT:
### Problem
The First-Time Setup → Terms & Conditions prompt (Do you agree to these terms? [yes/no]) re-appeared on every ./run.py, even on machines that were already set up and had previously accepted the terms.

Root ### cause
The prompt is gated by is_first_time_setup(), which returns not os.path.exists(PREFS_FILE_PATH) (.tt_studio_preferences.json). But nothing on the normal run path actually writes that file — it's only created by a handful of prompts (configuration_mode, npm_install_*) that don't fire on a default run. So .tt_studio_preferences.json was typically never created, is_first_time_setup() stayed True, and the terms prompt fired every time.


jashansinghTT (commented):
Terms didn't pop up when --dev flag was ran. It also didn't show up when I was --dev after the --cleanup flag.

jashansinghTT (approved):

PR #848, opened 2026-06-04
hotfix: reuse preloaded QB2 weights under pinned release

stisiTT:
## What's going on

Some machines were imaged with model weights already baked on disk under a folder named for the old `qb2_launch` build. We've since pinned tt-studio to a numbered tt-inference-server release, which looks for those weights under a folder named for the *resolved model version* — so it doesn't find the preloaded copy and re-downloads tens of GB.

This creates a symlink from the version-named folder to the already-present preloaded one, so the existing weights get reused.

## Why this lives in inference-api/api.py (deploy time), not run.py

The folder version is **not** the repo `VERSION`. It's the **per-model, catalog-pinned `model_spec.version`** (e.g. Qwen3-32B → `0.14.0`, not `0.15.0`), and it varies by impl/device. A first attempt in `run.py` at startup guessed the global VERSION and produced the wrong name (`-v0.15.0`) while the deploy actually wanted `-v0.14.0`. Startup can't know the right token — it doesn't know the request's device/impl.

So the symlink is now created at **deploy time** in `api.py`, resolving the version with the *same* `get_runtime_model_spec(model, device, impl)` that `setup_host` uses. The names match by construction, for any model/impl/device. The run.py helper is reverted.

## Notes
- **Symlink, not copy** — no duplicated disk.
- **Safe / invisible when not needed:** does nothing if there's no preloaded folder, if the target already exists (never overwrites), or for models not on the host-volume allowlist (`host_volume_models.json`). Machines already upgraded past the ISO are unaffected.
- **Temporary:** `TEMP (QB2 workaround)` banner-marked helper + one call — delete to fully revert.

## Testing
Verified locally: links the version resolved from `model_spec.version` (`-v0.14.0`, not the repo VERSION), weights reachable through the link, idempotent, never clobbers a real dir, NOP without an override. End-to-end reuse to be confirmed on a QB2 machine.

### DIFF
--- a/inference-api/api.py
+++ b/inference-api/api.py
@@ -481,6 +481,32 @@ def _model_uses_preferred_host_volume(model_name: str) -> bool:
     return (model_name or "").strip().lower() in _HOST_VOLUME_MODEL_ALLOWLIST
 
 
+# ─── TEMP (QB2 workaround) — remove this fn + its call in the deploy path ──────
+def _stage_preloaded_version_symlink(model, device, impl, override_dir, root, job_id):
+    """Link volume_id_<impl>-<model>-v{model_spec.version} -> the preloaded
+    -vqb2_launch override dir, so `--host-volume <root>` reuse lands on the
+    preloaded data instead of re-downloading.
+
+    The version is resolved with the SAME get_runtime_model_spec(model, device,
+    impl) that run.py/setup_host use, so the symlink name matches what setup_host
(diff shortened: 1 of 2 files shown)


hous:
I tested with preloaded Qwen3-32B, this works as expected with this symlink.

hous (approved):
Tested on QB2 with preloaded Qwen3-32B cache directory and it worked.

PR #868, opened 2026-06-08
Fix Deployment History showing N/A / Still running for stopped models

### LINKED ISSUE #863: Fix the deployment for the models in the model deployment page
Image

The models show, still running even though they have stopped. Also time stopped shows N/A

jashansinghTT:
The Deployment History page kept showing "N/A" for Stopped At and "Still running" for Duration even after a model had stopped.

The cause was that two backend paths marked a deployment as stopped without recording stopped_at: the stale-deployment reconcile in get_canonical_deployments and the failed-job handler in deployment_sync. This sets stopped_at in both places so the page shows the real stop time and duration.

This was the issue here in the form of an image:

image


anirudTT (approved):
lgtm

PR #872, opened 2026-06-08
Refactored all stop and reset views to unified streaming views

### LINKED ISSUE #852: refactor : reset views in delete model button vs reset button in navbar

rnabeelTT:
### Summary
Collapses the duplicated stop/reset views into a single streaming (SSE) stop view
and one streaming reset view, and surfaces live `tt-smi` logs across all destructive
flows (per-model delete, per-chip reset, reset-all). Previously the per-model delete
streamed over SSE while reset-all/per-chip used a separate JSON stop view plus
buffered `text/plain` blobs with no live output.

### Backend (`docker_control`)
- **One stop view** — `StopStreamView` (SSE) is now the only stop/delete path. Stops +
  removes the container, then resets its chips unless `?skip_device_reset=true`. The
  old JSON `StopView` is removed.
- **One reset view** — `ResetStreamView` (SSE) replaces the `text/plain` `ResetBoardView`
  and `ResetDeviceView` (board when no `device_id`, single chip otherwise).
- Shared async helpers (`_astream_stop_remove_container`, `_astream_tt_smi_reset`,
  SSE framing) back both views, so the streaming `tt-smi -r` logic lives in one place.
- Routes: `stop/stream/<container_id>/` (unchanged) + `reset_board/stream/`,
  `reset_device/stream/<device_id>/`. Removed dead routes/imports and `StopSerializer`.
- Performance preserved: reset-all still stops all containers in parallel with
  `skip_device_reset` and runs a single whole-board `tt-smi -r`; cache-refresh behavior
  matches the old per-endpoint logic.
- Stopping or deleting a model that is **still deploying** now sticks: the
  deployment-sync paths (`_do_sync`, `_sync_chat_deployment_record`) honor
  `stopped_by_user` and clean up the container the deploy job creates instead of
  resurrecting the row when the job completes.

### Frontend
- `deleteModel()` now drives the stop stream (resolves on `complete`); its signature
  and call sites are unchanged, so reset orchestration/parallelism is identical.
- Reset dialogs (`MultiCardResetDialog`, `ResetIcon`) consume the reset stream instead
  of reading a blob; added a shared `consumeSSE`/`streamResetAction` and a
  `StreamingLogPanel` so reset flows render live `tt-smi` output.
- Multi-chip models now render as a **single grouped card** in the device-reset
  grid (spanning their chips) with one Reset that stops the model and resets all
  its chips together — a multi-chip model's chips can no longer be reset
  individually (which previously stopped the model while leaving a chip un-reset).

### Tests & docs
- Updated backend integration tests to stop via the SSE stream.
- Updated the stop-endpoint docs in `README.md`.

### Test plan
- Delete an individual model and confirm live progress logs appear.
- Do a delete-all / reset-all and confirm live progress logs appear.
- Verify live logs show on both pages.
- Reset a model that occupies multiple chips → it shows as one card with one
  Reset, and both chips reset together (single `tt-smi -r 0 1`).
- Reset (or delete) a model while it is still deploying → its row disappears and
  does not reappear after the job finishes or after navigating away and back.

### UI Changes
Screenshot 2026-06-08 at 3 21 34 PM
Screenshot 2026-06-08 at 2 41 53 PM
Screenshot 2026-06-08 at 2 44 50 PM


anirudTT (approved):
lgtm

PR #876, opened 2026-06-08
Fix API examples returning 404 for non-LLM models

### LINKED ISSUE #861: API button : endpoints wont work for non LLM

jashansinghTT:
## Problem

Every deployed model has an **API** button that opens the API info page and shows ready-to-run curl commands. For non-LLM models — speech-to-text (Whisper, Distil), text-to-speech, image generation, object detection — the page generated the **same chat/completions curl** as LLMs. Running that command returned **404 Not Found**, because those model containers don't serve `/v1/chat/completions`.

## Fix

The API info now adapts to the model type:

- **Backend** (`ModelAPIInfoView`): for non-LLM models, returns a relative `inference_route` pointing at the matching TT-Studio proxy route (`/models-api/speech-recognition/`, `/tts/`, `/image-generation/`, `/object-detection/`). These are the same routes the rest of the UI already uses, so they return a real response and handle container auth/payload internally. Chat/VLM models are untouched.
- **Frontend** (`ExamplesTab`, `EndpointsTab`): when a model has an `inference_route`, the page builds a working curl against `window.location.origin + inference_route` with the correct payload per type — multipart `file`/`image` for speech-to-text/object-detection, JSON `text`/`prompt` for TTS/image generation — and shows that route as the endpoint. LLM pages are unchanged.

Why a relative route + frontend origin: behind the dev/prod proxy the backend can't see a host-reachable URL (it only sees the internal container name), but the frontend always knows the origin the user is actually on.

## Testing

Verified on a QB2 (Blackhole P300x2) machine with the dev stack running.

- **Speech-to-text (Whisper):** deployed `whisper-large-v3`, ran the generated curl
  ```
  curl -X POST "<origin>/models-api/speech-recognition/" -F "deploy_id=<id>" -F "file=@audio.wav"
  ```
  → `200 OK`, JSON transcription returned.
- **Text-to-speech (SpeechT5):** deployed `speecht5_tts`, ran the generated curl
  ```
  curl -X POST "<origin>/models-api/tts/" -H "Content-Type: application/json" -d '{"deploy_id":"<id>","text":"Hello from Tenstorrent"}' --output tts_output.wav
  ```
  → `200 OK`, valid WAV audio returned.
- **API info payload:** confirmed `inference_route` is set per `model_type` and the old chat-only output is unchanged for LLMs.
- Frontend `tsc` type-check passes.

Image generation and object detection use the identical mechanism (same proxy routing + payload matched to the existing views); they were not booted live here (image-gen needs the whole board; the legacy YOLOv4 deploy path stalls independently of this change). Worth a quick manual check once those models are up.

## Files changed
- `app/backend/model_control/views.py`
- `app/frontend/src/components/ModelAPIInfo/tabs/ExamplesTab.tsx`
- `app/frontend/src/components/ModelAPIInfo/tabs/EndpointsTab.tsx`
- `app/frontend/src/components/ModelAPIInfo/useModelAPIInfo.ts`


anirudTT (approved):
lgtm tested works for non llms

PR #893, opened 2026-06-09
Hotfix: Qwen device id issue fixed in the tt-smi blip case

### LINKED ISSUE #895: Device ID bugs in tt-studio
The bug
Multi-chip models (e.g. Qwen3-32B) intermittently recorded device_ids=[0] instead of [0,1,2,3], showing "Device 00" in the Models Deployed table (while the chip diagram, which recomputes live, stayed correct). The deploy-time code clamped the occupied-slot set with min(device_id + chips_required, allocator.total_slots), and total_slots is derived from board detection via tt-smi; when a tt-smi probe timed out (e.g. device contention, or a cold cache right after a backend restart), detect_board_type() returned "unknown", collapsing total_slots to 1 and baking [0] into the persisted deployment record.

rnabeelTT:
### The bug
Multi-chip models (e.g. Qwen3-32B) intermittently recorded device_ids=[0] instead of [0,1,2,3], showing "Device 00" in the Models Deployed table (while the chip diagram, which recomputes live, stayed correct). The deploy-time code clamped the occupied-slot set with min(device_id + chips_required, allocator.total_slots), and total_slots is derived from board detection via tt-smi; when a tt-smi probe timed out (e.g. device contention, or a cold cache right after a backend restart), detect_board_type() returned "unknown", collapsing total_slots to 1 and baking [0] into the persisted deployment record. 

### The fix
The fix derives the occupied set purely from the static catalog value (range(device_id, device_id + chips_required)), removing the dependency on flaky board detection so the recorded devices are deterministic and correct regardless of tt-smi state.

Fixes #895


anirudTT (approved):
lgtm.

PR #899, opened 2026-06-10
Hotfix: No pill selected in progress bar during downloading weights stage of cached LLMs

### LINKED ISSUE #900: Progress bar transitions from weights downloading to cached
### The Bug
When a model is not cached, the progress bar clearly shows that the weights are being downloaded. However, after the download is complete, it shows the "Cached" message it should only show for models that had their weights cached before the current run, and it also gets rid of the Downloading Weights Pill, and essentially no pill is selected for a few seconds.

rnabeelTT:
### Changes
- If the model is cached, the progress bar never shows no pill selected, it simply selects the next pill after the weights download stage.
- When weights get downloaded, the progress bar no longer shows cached as it did before. It only shows cached for the models with weights existing before the deployment run. Fixes #900 
- Removed an unnecessary message from the reset-all modal telling user that the board was already being reset even if the user was on the same screen constantly.


jashansinghTT (approved):
The progress bar didn't stop after the cache models, the qwen 32b was deployed with the cached version in this case. Along with this, the model showed up properly and accurately. The pills and the progress were appropriate in this case.

PR #904, opened 2026-06-10
Hotfix: Resolve Llama-70B deployment on P300X2 - forward override_docker_image on the chat deploy path

### LINKED ISSUE #903: Llama-70B fails to deploy on P300X2
**Description:**
Deploying Llama-3.3-70B-Instruct (or Llama-3.1-70B-Instruct) on a P300X2 board via TT-Studio fails. The TT Inference Server is handed the `model_spec` default image for 70B@P300X2 — `vllm-...:0.10.0-555f240-22be241` — and `run.py` (v0.15.0) rejects it:

> ⛔ Image v0.10.0 is not supported in this version of run.py (need v0.11.0+)

The v0.15.0 release changed the vLLM container's entrypoint contract and refuses any image below v0.11.0. The 8B deploys fine on the same board only because the server's `model_spec` already maps 8B@P300X2 → 0.14.0 (above the floor); the 70B's only P300X2 image in the spec is the old 0.10.0.

**Cause:**
TT-Studio already has logic to force the P300X2-compatible 0.14.0 image for these Llama models — but it lives in `run_container()` (`docker_control/docker_utils.py`). CHAT models (all the Llama LLMs) deploy through `start_chat_deployment()` via the `ModelTypes.CHAT` branch in `views.py`, which returns before `run_container()` is ever reached. So the override is ineffective for the models it targets — it's never included in the `/run` payload for any Llama model. The 8B happened to work because the server defaults it to 0.14.0 anyway; the 70B is the one case where the override is actually required, and it never fires.

**Fix (high level):**
Move the image override onto the CHAT deploy path so it's actually sent in the `/run` request. The inference server's `/run` endpoint accepts an `override_docker_image` field (present in its OpenAPI `RunRequest` schema), and `run.py`'s `--override-docker-image` re-parses the image version, clearing the v0.11.0 floor.

*Note: the proper long-term fix is server-side — repoint the `model_spec` P300X2 rows for the 70B models to a ≥v0.11.0 image so no override is needed. This TT-Studio change is the client-side workaround until then.*

---

rnabeelTT:
**Summary:**
Llama-70B failed to deploy on P300X2 because the inference server defaults 70B@P300X2 to a v0.10.0 vLLM image, which `run.py` (v0.15.0) rejects (needs ≥v0.11.0). TT-Studio had an override to force the P300X2-compatible 0.14.0 image, but it was set in `run_container()` — a path CHAT models never take (they deploy via `start_chat_deployment()`), so it never took effect. This PR moves the override onto the chat deploy path so it's actually sent, and removes the dead copy.

Validated on a P300X2 host: with the 0.14.0 override now in the `/run` payload, the 70B clears the version gate, loads weights, and initializes on the P300X2 mesh — matching a known-good manual `run.py --override-docker-image …` deploy.

**Changes:**
* **docker_control/tt_inference_client.py** — `start_chat_deployment()` gains an `override_docker_image` parameter and includes it in the `/run` payload when set (alongside the existing optional fields).
* **docker_control/views.py** — the CHAT deploy path computes the override image for the Llama models that need it, threads it into `chat_deploy_kwargs` (→ `start_chat_deployment`), and uses it as the pre-pull target (`deploy_image = override_docker_image or resolve_deploy_image(...)`) so the UI pre-pulls the image that will actually run. No-op for the 8B (server already resolves it to 0.14.0) and for non-Llama CHAT models (`override_docker_image` stays `None`).
* **docker_control/docker_utils.py** — removed the now-redundant override block from `run_container()` (it was unreachable for CHAT models) and left a note pointing to the new location on the chat path.

**Test Plan:**
- Launch tt-studio.
- Deploy Llama-3.3-70B-Instruct.
- Validate that the model is successfully deployed.

Fixes #903

### DIFF
--- a/app/backend/docker_control/docker_utils.py
+++ b/app/backend/docker_control/docker_utils.py
@@ -369,16 +369,6 @@ def run_container(impl, weights_id, device_id=0, host_port=None, use_image_overr
         # if use_image_override and impl.model_name in {"whisper-large-v3", "speecht5_tts"} and board_type == "P300x2":
         #     payload["override_docker_image"] = "ghcr.io/tenstorrent/tt-media-inference-server:qb2_launch-6900b0c-dev"
 
-        # These models use v0.14.0 image (P300X2 compatible)
-        if impl.model_name in {
-            "Llama-3.1-8B",
-            "Llama-3.1-8B-Instruct",
-            "Llama-3.1-70B",
-            "Llama-3.1-70B-Instruct",
-            "Llama-3.3-70B-Instruct",
-        }:
-            payload["override_docker_image"] = "ghcr.io/tenstorrent/tt-inference-server/vllm-tt-metal-src-release-ubuntu-22.04-amd64:0.14.0-80180b9-7678b70"
-
         logger.info(f"API payload: {payload}")
 
         # Make POST request to TT Inference Server API
--- a/app/backend/docker_control/tt_inference_client.py
+++ b/app/backend/docker_control/tt_inference_client.py
@@ -71,6 +71,7 @@ def start_chat_deployment(
     dev_mode: bool = False,
     skip_system_sw_validation: bool = True,
     override_tt_config: Optional[str] = None,
+    override_docker_image: Optional[str] = None,
 ) -> TTInferenceRunResult:
     """Start a chat model deployment via TT Inference Server (/run).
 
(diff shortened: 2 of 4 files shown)


anirudTT (approved):
lgtm , in terms of code diff , and will test and deploy on the rc, once its cherry-picked

anirudTT:
@stisiTT : for visibility and knowledge sharing

PR #907, opened 2026-06-10
Added preflight check for existance of enough chips before deploying VA

### LINKED ISSUE #909: VA : do not deploy unless all 4 chips are available

rnabeelTT:
Adds a preflight check to the Voice Agent solution screen that verifies the board exposes enough chip slots before deploying.

The pipeline pins fixed device slots (LLM/Whisper/TTS on 0/1/2, or 0,1/2/3 for the P300x2 Llama card-pair), but nothing validated that those slots actually exist. On an under-provisioned or undetected board, the deploy would proceed and fail with cryptic per-card "Invalid device_id" 409s.

Now we fetch total_slots from /docker-api/chip-status/, compare against the layout's requirement (3 or 4), and disable Deploy with a clear banner when the board reports too few — also flagging that the hardware may simply not be detected. The check is null-safe: if the chip-status probe fails, it stays out of the way.

Also fixed the strobing/flashy VA deployment UI to one that is more gentle and easy on the eyes.


anirudTT (approved):
lgtm

jashansinghTT (commented):
This worked for me properly. The slots shown were appropriate and in the UI as well. It showed the slots that were occupied and basically showed that all the slots were occupied.

jashansinghTT (approved):
This is good to merge.

PR #905, opened 2026-06-10
Refactor Chip and Device Terminology Across Components

anirudTT:
Updated terminology from "chip" to "device" in various components to enhance clarity and consistency. This includes changes in user interface texts, variable names, and comments. The refactor aims to align the language used throughout the application with the intended hardware configuration, improving user understanding and experience.

PR #911, opened 2026-06-11
Clarify Llama 3.1 8B Instruct in P300x2 and voice pipeline text

jashansinghTT:
The base `Llama-3.1-8B` (completions) model was removed from the model catalog, leaving only `Llama-3.1-8B-Instruct`. This updates the remaining user-facing strings that referred to a bare "Llama 3.1 8B" / "Llama 8B" so they name the Instruct variant — the only 8B model these flows can actually deploy.

## Changes
- **SelectionSteps**: P300x2 device-pair toast and the full-board flow comment now reference the Instruct variant.
- **VoiceAgentSolutionStep**: voice pipeline description now says "Llama 8B Instruct".

## Note
The chip-selection example tags in `ChipConfigStep` that previously listed the base `Llama-3.1-8B` (visible in the "Advanced: Configure Hardware" / chip-selection flow) are no longer present after the chip/device terminology refactor (#905), which replaced model-name example tags with generic descriptions. So no change is needed there — that part of the original issue is already resolved on `dev`.

## Verification
- `tsc --noEmit` passes clean.

### DIFF
--- a/app/frontend/src/components/SelectionSteps.tsx
+++ b/app/frontend/src/components/SelectionSteps.tsx
@@ -249,14 +249,14 @@ export default function StepperDemo() {
       const pair = pickPreferredAvailablePair(chipStatus?.slots);
       if (!pair) {
         customToast.error(
-          "Llama 3.1 8B on P300x2 needs both devices 0 and 1 free."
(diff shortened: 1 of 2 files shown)


anirudTT (approved):
lgtm

Issue #902, opened 2026-06-10
 reset view connection failure 


anirudTT:
[tt-studio-logs-ttbr-fef99060580d.zip](https://github.com/user-attachments/files/28805743/tt-studio-logs-ttbr-fef99060580d.zip)

## Bug Report

**Date:** 6/10/2026 5:36:47 PM
**URL:** http://localhost:3000/
**ZIP / diagnostics reference:** `ttbr-fef99060580d`

Match this issue to the diagnostics bundle: the downloaded ZIP from TT-Studio uses this reference in its filename. Paste it in a follow-up comment if you rename the file.

### Description
_Not provided_

### Steps to Reproduce
_Not provided_

### Expected Behavior
_Not provided_

### Actual Behavior
_Not provided_

---

TT-Studio diagnostics were collected separately and intentionally are not embedded in this GitHub issue to keep the report short and avoid oversized issue URLs/bodies.

Collected diagnostics include backend and tt-inference-server logs, deployment history, inference artifacts, tt-smi hardware data, and a current deployed-models snapshot (see current_models.json in the ZIP).

**Please attach the diagnostics ZIP to this GitHub issue** after you create it: in TT-Studio use **Download Logs as ZIP** (step 3), then on the GitHub issue attach that file (GitHub does not accept large bodies via URL).

*Auto-generated by TT-Studio Bug Reporter*

anirudTT:
- doesn't happen in the PROD mode

PR #915, opened 2026-06-11
Fix reset view connection failure: heartbeat the reset SSE stream

jashansinghTT:
## Problem

Issue #902 ("reset view connection failure"). A board reset reported as a
connection failure in the UI, even though the reset **succeeded on the hardware**
(the diagnostics bundle shows `Device state set to RESETTING` → `Device state cache
cleared`, ~42s, then a healthy board).

## Root cause

The board/device reset streams `tt-smi -r` output to the browser over Server-Sent
Events (`ResetStreamView` → `_astream_tt_smi_reset` → `_astream_reset_phase` in
`app/backend/docker_control/views.py`). The stream only sends bytes when `tt-smi`
prints a line. During the reset, `tt-smi` is silent for long stretches (~40s) while
it re-initializes the boards, so the SSE connection sits idle.

The frontend consumes the stream with a raw `EventSource` (`consumeSSE`,
`app/frontend/src/api/modelsDeployedApis.ts`) that rejects on the **first**
`onerror` with `"Connection to stream lost."`. If any intermediary or the browser
drops the idle connection during the silent window, the user sees a connection
failure even though the reset itself completed.

## Fix

Emit an SSE keepalive comment (`:keepalive\n\n`) at least every 15s while `tt-smi`
is quiet, so the connection is never idle long enough to be dropped. SSE comment
lines are ignored by `EventSource`, so the frontend protocol is unchanged.

The stall-detection behavior is preserved: only after the full `line_timeout` of
**continuous** silence does it escalate to terminating the process. The change
applies to both the whole-board reset and the per-device reset performed after a
model stop (both go through `_astream_reset_phase`).

Minimal, backend-only — one file, two functions:
- `_astream_tt_smi_reset`: wait for output in short (15s) intervals, emit a
  `heartbeat` on each quiet interval, accumulate toward the existing `line_timeout`
  stall threshold.
- `_astream_reset_phase`: translate `heartbeat` into an SSE comment line.

## Verification

- Verified on Tenstorrent hardware: a real board reset driven by a real (Chromium)
  browser EventSource streams `tt-smi -r` output and resolves successfully.
- Confirmed the full SSE path (Django `StreamingHttpResponse` → uvicorn → nginx/Vite
  proxy → browser EventSource) tolerates long silent windows.
- Unit-validated the new loop with a fake `tt-smi`: keepalives interleave during
  silent stretches, real output still streams, the stream completes, and the stall
  timeout path is preserved.

This is a defensive hardening: it keeps the reset SSE connection continuously active
so an idle drop cannot turn a successful reset into a "Connection to stream lost."
in the reset view.

### DIFF
--- a/app/backend/docker_control/views.py
+++ b/app/backend/docker_control/views.py
@@ -2439,6 +2439,7 @@ async def _astream_tt_smi_reset(device_ids, *, force_refresh):
     label = ", ".join(id_args) if id_args else "board"
     command = " ".join(["tt-smi", "-r", *id_args])
     line_timeout = max(90, 30 * len(device_ids))  # allow longer gaps on bigger boards
+    heartbeat = 15  # emit a keepalive at least this often while tt-smi is quiet
 
     await asyncio.to_thread(SystemResourceService.set_resetting_state)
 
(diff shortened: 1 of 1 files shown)


anirudTT:
Closing for now , this issue will be planned in the next sprint

PR #938, opened 2026-06-15
Hide Voice Agent solution on hardware without all required models (P100)

jashansinghTT:
## What

The Voice Agent "Solutions" deploy flow needs a board-compatible LLM, STT and TTS model. On hardware that can't run all three — e.g. **P100**, where only the LLM (`Llama-3.1-8B-Instruct`) is compatible and Whisper/SpeechT5 are not — the "Solutions" card is now hidden and the deploy screen goes straight to single-model deployment.

Gating is dynamic, driven by per-board compatibility from `get_containers` (`model_type` × `is_compatible`), so any board lacking one of the three model types collapses to single-model deploy.

## Files

- `app/frontend/src/components/SelectionSteps.tsx` — fetch the model catalog, compute `voiceAgentAvailable` from per-board compatibility, and render single-model deploy directly when it's false.

## Test plan

- [x] On a real P100, `get_containers` reports only `chat` compatible (`speech_recognition` + `tts` incompatible) → no Solutions card, no mode chooser, lands on single-model deploy. `?view=solution` deep-link also collapses to single.
- [x] Frontend typechecks clean (`tsc --noEmit`).
- [x] Regression: on P300x2 (all three compatible) the chooser + Solutions card still appear.

## Notes (not in this PR)

- The P100 deploy docker image is unchanged: `…:0.14.0-80180b9-7678b70` is correct — it satisfies run.py's `MIN_SUPPORTED_IMAGE_VERSION` (≥ v0.11.0) and exists on GHCR. (The release-spec's P100 image, v0.10.0, is rejected by that gate.)
- A working P100 deploy still requires the inference-api process to have docker-group access (it's launched without it when `run.py` runs from a shell that predates the `usermod -aG docker`), and the container's `CACHE_ROOT`/weights env to be provisioned. Both are environment/installer concerns.

### DIFF
--- a/app/frontend/src/components/SelectionSteps.tsx
+++ b/app/frontend/src/components/SelectionSteps.tsx
@@ -51,6 +51,10 @@ export default function StepperDemo() {
   const [totalSlots, setTotalSlots] = useState<number | null>(null);
   const isMultiChipBoard = totalSlots !== null && totalSlots > 1;
 
+  // Model catalog (with per-board compatibility) — used to decide whether the
+  // Voice Agent solution is even deployable on this hardware. null = not loaded yet.
+  const [models, setModels] = useState<Model[] | null>(null);
+
   // For QB2 boards, hide the hardware config step by default and show a toggle instead.
   const isQB2 = chipStatus !== null && QB2_BOARD_TYPES.has(chipStatus.board_type);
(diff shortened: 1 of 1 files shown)


anirudTT:
@anirudTT : to test this on the p100 machine

anirudTT (approved):

PR #962, opened 2026-06-17
Auto-install frontend npm deps on dev container start

jashansinghTT:
## Problem

In dev mode the frontend container's `node_modules` drifts out of sync with the live-mounted `package.json`. When a developer adds a new npm dependency, the running dev container never installs it, so imports fail until a full image rebuild or a `down -v` cleanup followed by a manual install.

### Root cause

- `app/frontend/Dockerfile.dev` runs `npm install` only at **image build time**, then the container command was just `npm run dev` (Vite), which never installs.
- `docker-compose.dev-mode.yml` bind-mounts `./frontend:/app` (live source incl. `package.json`) plus an **anonymous volume** `/app/node_modules`.
- Docker seeds that anonymous volume from the image **only once**, when it is first created, and never re-seeds it — even when the image is rebuilt. So a dependency added to the live `package.json` never reaches the running container.

## Fix

Override the dev frontend service command so it reconciles `node_modules` with the live `package.json` on every start, then launches Vite:

```yaml
command: sh -c "npm install --no-audit --no-fund --no-package-lock || echo 'npm install skipped/failed; using cached node_modules'; npm run dev"
```

- `npm install` is idempotent — a fast no-op when nothing changed, installs only the delta when a dep is added, and prunes deps removed from `package.json`.
- `--no-package-lock` keeps a routine dev start from rewriting the bind-mounted `package-lock.json`, so booting the stack no longer produces non-deterministic lockfile churn in the working tree. Interactive `npm install <pkg>` (run without this flag) still writes back to the host `package.json`/lockfile and is picked up on the next start.
- `|| echo ...` makes the install best-effort: if the npm registry is unreachable (offline, transient outage), the container falls back to the already-cached `node_modules` in the anonymous volume and still starts Vite, instead of the dev server failing to come up. Previously the install was baked into the image, so a built dev image could always start offline; this preserves that resilience.
- Scoped to dev mode only; prod compose and `Dockerfile.dev` are unchanged.

## Testing (on Tenstorrent hardware)

- Reproduced the bug: a dep added to `package.json` was missing in the container under the old command.
- After the fix, recreating the frontend auto-installed the missing dep (`added 1 package`) and Vite started; `http://localhost:3000` returned HTTP 200.
- Verified the in-container `npm install <pkg>` write-back to the host `package.json`.
- Verified idempotency (`up to date` fast no-op on warm restart) and pruning (`removed N packages` when deps are taken out).
- Verified the dev server still starts from the cached `node_modules` when `npm install` cannot reach the registry.
- Confirmed a routine start no longer modifies the host `package-lock.json`.
- All temporary test packages were removed; the only change in this PR is the one-line command override.


anirudTT:
closing for now , will re open later

anirudTT (approved):
Approving — this is at the right layer. The dev frontend live-mounts ./frontend:/app and keeps node_modules in an anonymous volume, and Docker only seeds that volume once, so a build-time npm install can never pick up deps added to the live package.json without a down -v + rebuild. Reconciling at container start is the only point where the live package.json and the persistent node_modules volume meet, so this solves the actual root cause.

PR #964, opened 2026-06-17
Fix cold-TTFT on first chat after idle

### LINKED ISSUE #688: Warm up before user can hit chat

### LINKED ISSUE #952: Inaccurate reporting of TTFT in chat UI
Image

The TTFT for the first chat is less even though it was noticeably slower than the second.

jashansinghTT:
## Summary
- vLLM (uvicorn) closes idle HTTP keepalive connections after 5s by default, so the first chat request after a brief idle pays a TCP+HTTP reconnect that inflates TTFT — first chat felt noticeably slower than subsequent ones, but the reported TTFT looked artificially low because the server-side timer starts after the connection is established.
- Adds a per-worker background warmer that runs **two warmup layers** through the same `_vllm_client` connection pool the inference path uses:
  1. `GET /health` every **500ms** — keeps the pooled TCP+HTTP keepalive socket genuinely active (well under uvicorn's 5s timeout).
  2. `POST /v1/{chat/}completions` with `max_tokens=1` every **30s** — warms vLLM's internal pipeline state (scheduler, allocator, tokenizer cache, prefill kernels) on top of the TCP layer. Fire-and-forget so it can never starve the /health cadence.
- Bumps `httpx` `keepalive_expiry` to 1h so the client is never the closer; the server stays the only closer, which the warmers prevent.

## Implementation
- New file: `app/backend/model_control/connection_warmer.py` — background thread submits `_vllm_client` coroutines onto the captured ASGI loop via `asyncio.run_coroutine_threadsafe`.
- Loop is captured lazily on first inference inside `stream_response_from_external_api` (idempotent, ~free after first call).
- `model_control/apps.py:ready()` starts the warmer thread; runs once per uvicorn worker.

## Test plan
- [x] Start the backend with a deployed model, confirm `connection_warmer: thread started (health_interval=0.5s, inference_interval=30.0s)` log appears per worker.
- [x] Send a first chat → confirm `captured ASGI loop` log appears.
- [x] Wait ~30s idle, send another chat → measured ~5x speedup on first-chat TTFT vs `dev`.
- [x] Confirm periodic `POST /v1/chat/completions` (max_tokens=1) requests appear in vLLM logs every ~30s and normal chats are unaffected.
- [ ] Reviewer: optionally repeat after multi-minute idle to confirm sustained warm behavior.

Fixes #952

### DIFF
--- a/app/backend/model_control/apps.py
+++ b/app/backend/model_control/apps.py
@@ -20,3 +20,6 @@ class ModelControlConfig(AppConfig):
         logger.info("Initializing models API")
         for model_id, impl in model_implmentations.items():
             impl.setup()
+
+        from model_control.connection_warmer import start_connection_warmer
+        start_connection_warmer()
--- a/app/backend/model_control/connection_warmer.py
+++ b/app/backend/model_control/connection_warmer.py
@@ -4,11 +4,16 @@
 
 """Periodic keepalive pings against deployed vLLM containers.
 
-vLLM (uvicorn) closes idle HTTP keepalive connections after 5s by default,
-so the first inference after a brief idle pays a TCP+HTTP reconnect on TTFT.
-A background thread submits cheap GET /health pings through _vllm_client
-(the same pool inference uses) onto the captured ASGI loop, keeping the
-pooled socket genuinely active.
+Two layers run from one background thread on each worker:
+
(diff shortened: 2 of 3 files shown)


anirudTT (approved):
lgtm tested on 8b model on qb2

Image

PR #967, opened 2026-06-18
Fix --cleanup-all reclaim size to include Docker volumes and images

jashansinghTT:
## What

`run.py --cleanup-all` reported an "Estimated disk to reclaim" and final "Reclaimed approximately X" that only counted host-side files (~200–300 MB). The command also removes the large stuff — Docker model-weight volumes (`volume_id_*`, tens of GB) and local images (vLLM, tt-studio/*, chroma) — but those were never measured, so the number was off by tens of GB.

## How

- Added `_docker_reclaimable_bytes()`, which reads `docker system df -v --format json` once (before the confirmation prompt, while the objects still exist) and sums the images matching `_CLEANUP_IMAGE_REFS`, the `volume_id_*` model volumes, and dangling anonymous volumes — the same objects cleanup-all deletes.
- Added `_parse_size_to_bytes()` to parse docker's SI size strings.
- Each Docker inventory line now shows its size, and the estimate + final total include them. Build cache is excluded (cleanup-all doesn't prune it). Degrades gracefully to the old host-only number if docker is unavailable.

## Validation

- `python3 -m unittest test_run_cleanup` — 18 tests pass (added 3, made the two `cleanup_all` tests hermetic).
- `printf 'n\n' | python3 run.py --cleanup-all` now shows ~73 GB (model volume 36.7 GB + images 35.5 GB + anon 0.6 GB) instead of ~300 MB, and still aborts before deleting anything.

### DIFF
--- a/run.py
+++ b/run.py
@@ -32,6 +32,7 @@ import platform
 import argparse
 import shutil
 import re
+import fnmatch
 import getpass
 import webbrowser
 import socket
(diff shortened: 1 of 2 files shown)


anirudTT (approved):
lgtm

PR #969, opened 2026-06-18
reset stream fixes and overhaul 

### LINKED ISSUE #902: reset view connection failure

rnabeelTT:
# Fix "Connection stream error" during Reset All + Re-enable Reset Board Button

## Summary
Fixed an intermittent Connection stream error during full device resets. The hardware reset (tt-smi -r) sends no data for long stretches while power-cycling chips, and several layers were closing the "silent" progress stream prematurely. Hardening the proxy/server timeouts and guaranteeing the backend always sends a terminal event makes resets reliable — which lets us safely re-enable the Reset Board navbar button.

---

## The Issue
The reset progress stream (SSE) was being torn down mid-reset by whichever layer had the shortest idle timeout:

- **Prod (Nginx):** The `/docker-api/` block had no timeout config, so Nginx used its 60s default and killed the connection after 60s of silence.
- **Dev (run.py --dev):** Dev serves the frontend via Vite (not Nginx) and proxies to uvicorn, which defaulted to a 5s keep-alive timeout — so the same crash happened far sooner in dev.
- **Abrupt backend closes:** An unhandled error inside the stream generator closed the response with no terminal event, which the client also reports as a lost connection.
- **Hidden UI:** The "Reset Board" button was hidden as a crude workaround for this flakiness.

---

## The Solution

- **nginx.conf:** Brought the `/docker-api/` block to parity with the stable `/models-api/` streaming config — `proxy_read_timeout 1200s` and `proxy_buffering off` so progress logs stream instantly and the connection survives the silent reset window. (Kept generous since `/docker-api/` also serves long deploys/image-pulls.)
- **docker-compose.dev-mode.yml:** Added `--timeout-keep-alive 1200` to the dev uvicorn command so dev matches prod.
- **docker_control/views.py:** The stop/reset stream generators now always emit a terminal `complete` event (even on an unexpected error), so the stream never closes abruptly.
- **modelsDeployedApis.ts:** Raised the client-side stream timeout from 180s → 300s, comfortably above the worst-case reset.
- **NavBar.tsx:** Removed the visibility guards so the Reset Board button is always visible and functional.

---

## Miscellaneous Improvements to Reset Flow

- **Hardware badge:** previously disappeared during a reset because the board status reads as unknown while `tt-smi` runs. The board identity is now retained, so the badge keeps showing the actual hardware.
- **Spurious timeout toast:** during a reset, the background `GET /docker-api/status/` poll stalls and triggered a "Request timeout: Server took too long to respond" toast. This polling is now paused while the device is resetting.

---

## Testing Plan

- Verify Reset All completes the full hardware power cycle without dropping the stream, in both `run.py` (prod) and `run.py --dev`.
- Confirm the Reset Board button is permanently visible in the UI.
- Confirm no spurious "Server took too long" toast appears during a reset, and the hardware badge stays visible throughout.


hous:
Appears to work as expected. Observed some strange behavior that is the result of the reset being a modal. Once the reset completes, it still appears that the voice solution is deploying. Is it still deploying?

Perhaps we can avoid this situation entirely by having reset be a separate page/view instead of making it possible to reset within every view?

image

anirudTT (approved):

PR #1002, opened 2026-06-29
Delete delayed reset in footer and premature closing of model deletion modal

rnabeelTT:
## Summary

Fixes the model deletion flow on the Models Deployed page, where deleting an individual model caused the progress modal to close prematurely and the board reset to appear to "restart" afterward.

## Background

The delete/reset progress modal (`DeleteModelDialog`) was rendered below the `rows.length === 0` early return in `ModelsDeployedCard`. Deleting a model marks it stopped immediately, so when its last model left the table the component switched render branches and **unmounted the modal mid-reset** — before the SSE stream's terminal `complete` event. The reset kept running in the background, which the global footer indicator then surfaced as a second, delayed "reset."

## Changes

- **Modal no longer closes early.** The dialog is hoisted to a stable top-level position rendered in both the populated and empty-state branches, so it survives the table collapsing to empty and stays open until the stream actually completes.
- **Footer updates instantly on "Run in background."** Minimizing now forces an immediate device-state refresh, so the footer's resetting indicator appears the moment the modal closes instead of waiting for the next scheduled poll.

No changes to the SSE protocol, close logic, or backend.

## Test Plan

1. Deploy a model and wait to reach Models Deployed page, not necessary to wait for status to become healthy.
2. Click **Delete** on the model and confirm.
3. Verify the modal **stays open throughout** both phases (stop container → reset devices) and does not flash, flicker, or close on its own — it should only close once the stream reports completion.
4. Repeat the delete, and this time click **Run in background** during the reset.
5. Verify the footer shows the **Resetting…** indicator immediately on close, not after a delay.
6. Confirm the model is gone and the board returns to a ready state afterward.

### DIFF
--- a/app/frontend/src/components/models/ModelsDeployedCard.tsx
+++ b/app/frontend/src/components/models/ModelsDeployedCard.tsx
@@ -51,6 +51,15 @@ import { useDeleteStream } from "../../hooks/useDeleteStream";
 import axios from "axios";
 import { ChipStatusDisplay } from "../ChipStatusDisplay";
 
+const deviceIdsForRow = (
+  row?: { device_ids?: number[]; device_id?: number | null },
+): number[] | undefined => {
+  if (!row) return undefined;
+  if (Array.isArray(row.device_ids) && row.device_ids.length > 0) return row.device_ids;
+  if (row.device_id != null) return [row.device_id];
+  return undefined;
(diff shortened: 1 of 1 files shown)


anirudTT (approved):
lgtm

PR #1004, opened 2026-06-29
updated to v0.17.0 artifact, fixed qwen image and device ids issue

rnabeelTT:
## Transition to v0.17.0 + Qwen deployment fixes

Moves the TT Inference Server artifact to the `v0.17.0` release and fixes two
Qwen deployment issues that surfaced with it.

### Changes

**1. Pin artifact to v0.17.0**
Switched from the temporary `TT_INFERENCE_ARTIFACT_BRANCH` commit SHA to
`TT_INFERENCE_ARTIFACT_VERSION=v0.17.0`.

**2. Fix Qwen failing to deploy on v0.17.0**
`run.py` was forcing `MODEL_SPECS_ENV=dev`, which made Qwen resolve to the wrong
(dev) model spec/image and fail to deploy on v0.17.0. Commented out the override
so deployments use the correct prod specs. (Block is marked with a TODO to remove
entirely once confirmed unneeded.)

**3. Fix multi-device models showing only "Device 00"**
When a model's image wasn't already cached, the deploy flow created a placeholder
record with `device_ids=device_ids` (just the base slot, `[0]`) instead of the
full `occupied_device_ids` set (`[0,1,2,3]`). The placeholder is only repointed —
never corrected — so the Models Deployed device column showed a single device for
multi-chip models like Qwen. Fixed the placeholder to store `occupied_device_ids`,
matching the cached-image path.

### Impact
- Qwen (and other v0.17.0 models) deploy correctly.
- Multi-device models report their full device set regardless of image cache state.
- Single-chip / explicit-slot deployments are unaffected.

### DIFF
--- a/.env.default
+++ b/.env.default
@@ -19,8 +19,8 @@ BACKEND_API_HOSTNAME=tt-studio-backend-api
 # NOTE: pinned to a commit SHA — the Wan2.2-T2V TT_DIT_CACHE_DIR dev-catalog fix lands
 # in commit b788278 (tenstorrent/tt-inference-server). The source branch was deleted but
 # the SHA is still valid. Re-pin to a release version once that fix ships in a tag.
-TT_INFERENCE_ARTIFACT_BRANCH=b788278762b0bcc9332ad453d265050bfa94738d
-# TT_INFERENCE_ARTIFACT_VERSION=v0.15.0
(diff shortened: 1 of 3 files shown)


anirudTT (approved):
lgtm deployed the qwen model .

anirudTT:
tested 8b as well

PR #1011, opened 2026-06-30
Fix hf cli command to the new hf download command

rnabeelTT:
### Old Command
`huggingface-cli download <model repository>` (legacy/deprecated)

### New command
`hf download <model repository>` (new)


anirudTT:
@hous thanks for pointing this fix .

anirudTT (approved):

PR #1017, opened 2026-06-30
Loosen download stall threshold and probe timeouts

stisiTT:
## Problem

Large model weight downloads were being **falsely flagged as stalled / timed out** during deployment, even while the download was healthy.

Root cause is a mismatch between our watchdog's impatience and the realities of a long `hf download` driven by the upstream tt-inference-server:

- The stall detector in `inference-api/api.py` flagged a job `stalled` after just **120s** without byte progress. That's shorter than several *legitimate* quiet gaps — HF shard verification, between-batch pauses, and slow `HEAD` requests — so healthy downloads got marked stalled.
- The progress probes that feed the liveness heartbeat used aggressive timeouts (`du` 4s, HF tree API 10s, log tail 3s). A transient probe miss froze the measured byte count, which then looked like a byte stall and compounded the false flag.

This was confirmed against a live deployment: a Qwen3-32B download reached 47 GB (15/27 files) then lost DNS inside the container. Byte progress is the only signal that correctly distinguishes that genuinely-dead download from a healthy-but-quiet one — the container stays `running` and the process stays alive in both cases — so the fix keeps byte-progress stall detection and tunes the threshold rather than removing it.

## Changes

- `inference-api/api.py`: `PULL_STALL_THRESHOLD_SECONDS` **120 → 600**. Ten minutes rides out normal quiet gaps while still surfacing a genuinely dead download within ~10 min, instead of waiting out the 5-hour hard timeout. (Comment updated.)
- `app/backend/model_control/download_progress.py`: `du` probe **4s → 15s**; HF tree API **10s → 30s**.
- `inference-api/api.py`: HF tree API **10s → 30s**.
- `app/backend/model_control/views.py`: log-tail probe **3.0s → 10.0s**.

This is the minimal, threshold-only hotfix. The existing 5-hour hard timeout remains as a backstop; `stalled` stays visible/actionable rather than being masked. A larger follow-up (zero-byte hard-fail, classifying the stall reason) was intentionally left out of scope.

## Verification

- `ast.parse` clean on all three modified Python files.
- Behavioral check: against the live Qwen3-32B stall (DNS dropped, ~51 min no byte writes), byte-progress detection still trips well before the 5-hour cap; healthy downloads with sub-10-min quiet gaps no longer false-flag.


anirudTT:
generally we avoid this workflow of pushing a PR to the rc , but this is a good fix . 

Can u push a PR to dev ill approve it ?

anirudTT:
@stisiTT — I think this is also still a valuable fix. Would you like to cherry-pick it and open it as a PR off dev?

PR #1019, opened 2026-06-30
Raise HF download timeouts to survive transient network blips

stisiTT:
## Problem

Model weight downloads inherited huggingface_hub's **10-second** default timeouts (`HF_HUB_ETAG_TIMEOUT`, `HF_HUB_DOWNLOAD_TIMEOUT`). A brief slow/flaky network patch on a metadata `HEAD` or a blob read therefore failed fast and burned through the retry budget, turning a recoverable hiccup into a frozen, multi-hour stall.

This was observed live: a Qwen3-32B deploy reached 47 GB then logged `read timed out (read timeout=10)` followed by name-resolution errors, and made zero byte progress for ~51 minutes while the process stayed alive.

This is the **root-cause** companion to the detection-side hotfix in #1017 (which only makes a dead download *visible* sooner).

## Change

Introduce `HF_DOWNLOAD_RESILIENCE_ENV = {HF_HUB_ETAG_TIMEOUT: 30, HF_HUB_DOWNLOAD_TIMEOUT: 60}` and propagate it to the actual `hf download` via the two paths tt-studio already controls:

- **In-container downloads (vLLM + media):** upsert the vars into the inference-server `.env` inside `sync_tokens_from_tt_studio()`, so they reach the container through the same `--env-file` path that already delivers `HF_TOKEN`. The upsert is idempotent and corrects stale values without duplicating lines.
- **Host-side `setup_host` downloads:** add the vars to `env_vars_to_set` (applied to `os.environ`).

All in `inference-api/api.py` (+30/−2).

## Why this is purely tt-studio

These are pure huggingface_hub runtime config — no package or image change. Verified the container's `--env-file` is the artifact `.env` that tt-studio already writes (the live container had `HF_TOKEN` delivered exactly this way).

## Scope decisions

- **Xet:** left disabled (`HF_HUB_DISABLE_XET=1`), unchanged — it was disabled deliberately for a correctness bug ("Xet exits 0 before blobs finish", commit `94438b9`). Not reopened.
- **hf_transfer:** rejected — the `hf_transfer` package is **not** in the tt-inference-server image (confirmed `ModuleNotFoundError` in the live container), so enabling it would break downloads and require an image change (not purely tt-studio).

## Verification

- `ast.parse` clean on `inference-api/api.py`.
- Upsert logic unit-checked in isolation: adds when missing, idempotent on re-run, corrects a stale/quoted value with no duplicate lines.
- **Decisive runtime check (to run on next deploy):** `docker inspect <container> --format '{{range .Config.Env}}{{println .}}{{end}}' | grep HF_HUB` should show both vars in the container env, the same way `HF_TOKEN` appears today — proving the env reaches the real `hf download`.

## Operational note

Takes effect on the **next** deploy: restart the inference-api service (to load this code) and redeploy the model (new container picks up the new `.env`). Existing partial weights resume — `snapshot_download` resumes by default and weights live in a persistent named volume, so no need to delete them.


anirudTT:
@stisiTT — I think this is still a valuable fix. Would you like to cherry-pick it and open it as a PR off dev?

PR #1021, opened 2026-07-03
Show model navbar entries only once the model is healthy

jashansinghTT:
### Problem
Per-model navbar entries (Chat UI, Image Generation, etc.) and the Voice Agent / Coding Agents links appeared as soon as a model was deployed — while the container was still warming up and couldn't serve requests.

### Change
Gate these entries on the model's real readiness. The navbar now probes `/models-api/health/` (the same signal `HealthBadge` and the models-table "Open" button already use) and only shows entries for models that report healthy. Deploying models are hidden until they're ready.

Frontend only, no backend changes.

### Verification
- `npm run type-check` and `npm run build` pass; lint clean on the changed files.
- Verified in the running dev app (HMR compiled cleanly). The health mapping (200 healthy / 202 starting / 503 unavailable) mirrors the existing `HealthBadge` path.


rnabeelTT (approved):
LGTM!

PR #1022, opened 2026-07-05
fix: use rsplit to strip full service_route in _video_base_url

### LINKED ISSUE #1000: Image-to-Video (I2V) models do not work — status/download URLs 404
## Summary
I2V (image-to-video) media models currently do not work at all. Every generation polls/downloads against a malformed URL and 404s. Surfaced during #835 testing.

## Symptom
- I2V deploy succeeds, but status polling and the final download both return 404; generation never completes.

## Root cause
- I2V models use service route `/v1/videos/generations/i2v`. The backend video proxy's `_video_base_url()` derives the poll/download base by stripping only the substring `/v1/videos/generations`, leaving a dangling `/i2v` on the base URL, so the resulting status/download URLs are malformed (`app/backend/model_control/views.py`, ~L763).
- The catalog side is correct: `map_service_route` returns `/v1/videos/generations/i2v` for I2V. The defect is purely in the proxy's base-URL derivation.

## Fix
- Strip the model's actual `service_route` suffix (e.g. `rsplit("/v1/videos/generations", 1)[0]`) instead of a substring `replace`, so the `/i2v` segment is removed cleanly.

## Status
Open. Blocks all I2V model deployments. T2V (Wan/Flux) is unaffected because its route has no extra suffix.

Refs: #835, #997.

abhi-0203:
## Summary

The `_video_base_url` function used `str.replace("/v1/videos/generations", "")` to derive the base URL from the `internal_url`. For I2V models whose `service_route` is `/v1/videos/generations/i2v`, this left a dangling `/i2v` segment on the base URL, causing all status polling and download URLs to 404.

## Root Cause

- I2V models use service route `/v1/videos/generations/i2v`
- `_video_base_url()` called `internal_url.replace("/v1/videos/generations", "")` 
- This only stripped the base prefix, leaving `/i2v` attached
- Result: poll URL became `http://host:port/i2v/v1/videos/generations/{job_id}` instead of `http://host:port/v1/videos/generations/{job_id}`

## Fix

Now uses `rsplit(service_route, 1)` to strip the full service_route suffix (including any model-specific path like `/i2v`), producing correct base URLs for both T2V and I2V models.

## Test Plan

- [x] Verified fix handles both `/v1/videos/generations` (T2V) and `/v1/videos/generations/i2v` (I2V) service routes correctly
- [x] No existing tests broken

Closes #1000

### DIFF
--- a/app/backend/model_control/views.py
+++ b/app/backend/model_control/views.py
@@ -746,7 +746,11 @@ def _video_base_url(deploy_id):
     """Resolve the tt-media-server base URL for a deployed video model."""
     deploy = get_deploy_cache()[deploy_id]
     internal_url = "http://" + deploy["internal_url"]
-    return internal_url.replace("/v1/videos/generations", "").rstrip("/")
+    service_route = deploy["model_impl"].service_route
+    # Use rsplit to strip the full service_route suffix (e.g. "/v1/videos/generations/i2v")
+    # instead of replace which only strips "/v1/videos/generations" and leaves a dangling
+    # suffix like "/i2v" on the base URL.
+    return internal_url.rsplit(service_route, 1)[0].rstrip("/")
 
 
 def _normalize_video_phase(raw_status):


anirudTT:
@abhi-0203 — Thanks for opening this PR!! and fixing the issue! Could you rework it off dev, cherry-pick your changes, and follow the [git branching guidelines](https://github.com/tenstorrent/tt-studio/blob/main/CONTRIBUTING.md)?

PR #1023, opened 2026-07-06
Show a blocking overlay when the backend is unreachable

jashansinghTT:
## Problem

When the Django backend becomes unreachable — most commonly an SSH port-forwarded session whose host machine sleeps or dies — the frontend keeps rendering every page. Users hit a scatter of low-level failures ("server not found", empty lists, stalled spinners, error toasts) with no single clear signal that the backend itself is gone.

## Change

- New `BackendHealthProvider` polls the backend's lightweight `/up/` liveness endpoint (~5s healthy, ~3s once a failure is seen, 5s request timeout). After 2 consecutive failures it renders a full-screen, non-dismissable overlay explaining the backend is disconnected, with recovery steps (restart the service, check the machine, check the SSH port forward) and a "Retry now" button. The overlay clears automatically once the backend responds again.
- The overlay renders via a portal above the toaster's z-index, so stray per-feature error toasts are covered rather than competing.
- Routed `/up/` through both the Vite dev proxy and the prod nginx config (it wasn't proxied before).

Provider is wired as the outermost provider in `routes/index.tsx`, below `ThemeProvider`, so the overlay covers the whole app and is themed correctly.

## Verification

- `npm run type-check`, `npm run lint` (new files clean), and `npm run build` pass.
- `curl` of `/up/` returns 200 directly (`:8000`) and through the Vite proxy (`:3000`).
- Stopped the backend container → proxied `/up/` returns non-200 (the failure signal that drives the overlay); restarted it → back to 200 (auto-recovery signal).

### DIFF
--- a/app/frontend/nginx.conf
+++ b/app/frontend/nginx.conf
@@ -65,6 +65,16 @@ server {
         proxy_set_header X-Forwarded-Proto $scheme;
     }
 
+    # Lightweight backend liveness probe used by the frontend to detect when the
+    # backend becomes unreachable.
+    location = /up/ {
+        proxy_pass http://tt-studio-backend-api:8000/up/;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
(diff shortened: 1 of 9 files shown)


anirudTT (approved):
lgtm tested by manually stopping the back-end .

PR #1030, opened 2026-07-08
Parallel model deployments before container start phase

### LINKED ISSUE #1006: Unclear workflow when deployment is in image pull stage
### The Problem
When the user hits deploy and a Docker Image is being pulled for vllm/media inference engine, if the user hits previous to go back, they can pick a different model, which still takes them to the same progress page. 

The overall flow is confused and a user wouldnt know which deployment is currently in progress until the image pull stage is over.

rnabeelTT:
### Summary
Reworks the single/multi-model deploy flow so users can deploy several models **in parallel** instead of being locked to one deploy at a time. Firing a deploy is now "fire-and-forget": the request is handed to a session-wide tracker, the button resets, and the user can go back, pick another model, and deploy it onto free devices while the first is still pulling its image. Progress is surfaced in a floating, minimizable banner, and device availability (including in-flight, still-pulling deploys) is reflected everywhere a model or device is selected.

### What changed in the workflow
- **Parallel deploys** — deploying no longer blocks the wizard. After a deploy is accepted, the user is free to navigate back and start another; multiple deploys run concurrently (the same behavior the Voice Agent flow already relied on).
- **Reservation-aware availability** — the moment a deploy is fired, its device(s) are reserved and greyed out in both the **model list** and the **advanced device configuration** menu — during the image-pull phase, not just after container start. Reservations are released automatically if the pull/start fails.
- **Resume the exact run** — reselecting a model that's already deploying reconnects to its live progress instead of offering a duplicate deploy.
- **Success → redirect preserved** — when the model being viewed finishes, it shows a success state and routes to Deployed Models.
- **Floating progress banner** — a subtle, minimizable bottom-right banner lists every in-flight/finished deploy with a compact progress bar; the detailed bar still shows inline within the deploy step.

### High-level code changes
- **`hooks/useActiveDeployments.ts` (new)** — session-wide source of truth for in-flight deploys: persists across navigation/reload, polls each job's progress independently, exposes `reservedDeviceIds`, and frees a reservation on terminal (failed) / refetches chip status on resolve.
- **`components/DeploymentTray.tsx` (new)** — fixed-position, minimizable banner rendering one compact row per deploy (progress bar, status, device, logs-on-failure).
- **AnimatedDeployButton, FirstStepForm, ChipConfigStep** — Streamlined the deploy button into a fire-and-forget trigger, and updated the form and configuration steps to grey out unavailable models and pending hardware devices.

### Testing plan
1. **Parallel deploy (auto mode):** Deploy an uncached model (e.g. `speecht5_tts`). While it pulls, hit **Prev** → confirm its device is greyed and models needing it are disabled with a reason. Select a model that fits the remaining free devices (e.g. Llama-8B → devices 2,3) and deploy → both pulls run concurrently in the banner.
2. **Parallel deploy (advanced config):** Repeat with the advanced device configuration open → confirm the pending device shows "IN USE" during the pull and the free devices/card remain selectable.
3. **Resume:** With a deploy in flight, reselect that same model → confirm the deploy step reconnects to the live bar rather than offering a fresh deploy.
4. **Completion:** Let the viewed model finish → success message shows and redirects to Deployed Models; other in-flight deploys keep running in the banner.
5. **Reload/navigation:** Refresh mid-pull → confirm active deploys resume from persisted state; navigate away and back → tracker state is intact.
6. **Regression:** Single sequential deploy on a single-chip board, cached-image deploy (no pull bar), and the Voice Agent flow all behave as before.

Closes #1006


jashansinghTT (approved):
This PR works at the present perfectly

PR #1062, opened 2026-07-13
fix(workflow): default tab and render markdown in node inspector

### LINKED ISSUE #1033: Workflows node inspector: put Output before Config + render output as markdown
Part of #1031

### Problem

In the node inspector, the **Config** tab is first and selected by default; users want **Output** first. The output is also rendered as raw text, so model/markdown output isn't formatted.

### Current behavior

`NodeConfigPanel.tsx` declares the Config tab (~L195) before the Output tab (~L206) and defaults `activeTab` to `"config"` (~L78). Output is rendered in a plain `<p ... whitespace-pre-wrap>` (~L164) — **not** markdown.

### Proposed

- Reorder tabs to **Output → Config**.
- Default to the Output tab when the node has output; fall back to Config otherwise.
- Render output through the app's existing markdown renderer (reuse the chat UI's markdown component rather than adding a new dependency).

### Acceptance criteria

- [ ] Output tab is first and shown by default for a node that has output.
- [ ] Headings / lists / code / tables render as formatted markdown.
- [ ] A node with no output still opens sensibly.

### Key files

- `app/frontend/src/components/workflows/NodeConfigPanel.tsx`

AnkittChauhan:
### Description

This PR addresses the UI/UX issues in the Workflows Node Inspector by updating the tab ordering and rendering the output as formatted markdown.
 - resolves #1033 

**Changes:**
1. **Reordered Tabs:** The `Output` tab is now placed before the `Config` tab.
2. **Dynamic Default Selection:** Added a `useEffect` to `NodeConfigPanel` that intelligently checks the store when a node is selected. If the node has generated output or reasoning logs, it defaults to the `Output` tab. For fresh nodes with no output, it safely falls back to `Config`.
3. **Markdown Rendering:** Swapped the raw `<p>` text tag for the app's existing `MarkdownComponent` so that LLM outputs (tables, code blocks, lists, etc.) are correctly formatted.

### Testing
Since deploying an actual LLM requires TT hardware which I do not currently have access to locally, I validated the markdown UI rendering by directly injecting a rich markdown payload into the Zustand `workflowStore`. 

As shown in the screenshot below, the Output tab now correctly defaults to active and perfectly renders complex markdown elements including tables and code blocks.

**Screenshots:**
Screenshot 2026-07-13 at 3 07 31 PM

### DIFF
--- a/app/frontend/src/components/workflows/NodeConfigPanel.tsx
+++ b/app/frontend/src/components/workflows/NodeConfigPanel.tsx
@@ -19,6 +19,7 @@ import InputConfigPanel from "./config/InputConfigPanel";
 import LLMConfigPanel from "./config/LLMConfigPanel";
 import RAGConfigPanel from "./config/RAGConfigPanel";
 import AgentConfigPanel from "./config/AgentConfigPanel";
+import MarkdownComponent from "../chatui/MarkdownComponent";
 
 interface SourceLink {
   title: string;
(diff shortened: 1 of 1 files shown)


jashansinghTT (approved):
Thanks for the fix. This works properly in the app and the markdown renders properly

Image

PR #1074, opened 2026-07-14
Remove sudo password prompt from setup

### LINKED ISSUE #1047: Remove sudo requirement from TT-Studio setup & runtime
## Summary

Remove TT-Studio's reliance on `sudo`. Setup and runtime currently shell out to `sudo` in several places, which is a friction/security concern and blocks locked-down environments. The goal is a `sudo`-free happy path (with clear guidance when the user genuinely lacks Docker access).

## Where sudo is used today

- `run.py` — `--no-sudo` flag ("Skip sudo usage for FastAPI setup") already exists; make sudo-free the default/robust path.
- `tt_setup/docker.py` — sudo fallbacks for `docker compose version`, `docker info`, and daemon reachability (retry with sudo for 660 sockets).
- `tt_setup/services/_docker_control.py` — `sudo kill` of the Docker Control process on port 8002; warns when it can't kill without sudo.
- `tt_setup/docker_diag/_diagnostics.py` — sudo in remediation hints (`usermod -aG docker`, `dmesg`, `chown`, rebuild commands) and `verify_docker_containers(use_sudo=...)`.

## Approach

- [ ] Prefer rootless/group-based Docker access: detect docker-group membership and use it instead of sudo.
- [ ] Replace `sudo kill` on port 8002 with a user-owned process-management path (track our own PID, kill only our own process).
- [ ] Where a 660-permission socket is detected, surface the one-time fix (`usermod -aG docker $USER` + re-login) as guidance instead of silently escalating.
- [ ] Make `--no-sudo` behavior the default; keep an explicit opt-in escape hatch only if truly required.
- [ ] Verify the full `run.py` flow works end-to-end with a non-sudo user in the docker group.

## Deliverable

`python run.py` completes with no `sudo` invocation for a user in the `docker` group; degraded environments get actionable guidance, not silent privilege escalation.

rnabeelTT:
### Problem
`python3 run.py` prompted for a sudo password on nearly every invocation. The prompt came from a proactive `request_sudo_authentication()` call in the inference-server setup, kept around so cleanup could later delete files that
model/app containers had written as root (or uid 1000) into the artifact and persistent-volume directories.

### Change
- **Drop the proactive sudo prompts.** Removed both up-front `request_sudo_authentication()` calls in `setup_tt_inference_server` (`tt_setup/inference_server/_orchestrator.py`). Nothing asks for a password on
  the common path anymore.
- **Clean up container-owned files with a container, not sudo.** Added `_remove_path_with_docker()` in `tt_setup/cleanup/_resource_ops.py`, which removes a path via an ephemeral `alpine` container using the caller's existing Docker access. `_remove_path`'s `PermissionError` fallback and `remove_artifact_with_sudo` now route through it, so cleanup operations works non-interactively regardless of who owns the leftover files.
- **Run the low-footprint services non-root.** Set `user: "1000:1000"` on`tt_studio_agent` and `tt_studio_litellm` in `app/docker-compose.yml` (neither writes to host bind mounts). Backend / frontend / chroma are intentionally left as-is — they need image-level changes and are out of scope here.

`request_sudo_authentication` is left defined (still re-exported from`tt_setup.inference_server`) but is now unused internally.

### Verification
- All changed modules byte-compile; `tt_setup.inference_server` imports cleanly.
- Confirmed separately that the cleanup container removes root-owned files that the host user cannot delete, with no sudo, and that agent + litellm start and stay healthy as uid 1000.

Closes #1047


jashansinghTT (approved):
This worked for me, no sudo password was required in this case.

anirudTT (approved):
lgtm , maybe take a quick look at the copilot comments

PR #1084, opened 2026-07-15
Anirud/reconfigure inf server scope

### LINKED ISSUE #1083: Scope --reconfigure-inference-server + verify refs upstream & recover from corrupt artifacts
## Summary

Make `python run.py --reconfigure-inference-server` do one focused thing well, and harden how the inference-server artifact is fetched. Work is on branch `anirud/reconfigure-inf-server-scope`.

## Changes

1. **Scope the reconfigure flow.** `--reconfigure-inference-server` now prompts only for the artifact source (release version vs. branch) and skips the full environment sweep (HF_TOKEN, secrets, ports, …). A full `--reconfigure` still runs the entire flow.
2. **Verify refs upstream.** Reject a branch / release tag / commit SHA that does not exist in `tenstorrent/tt-inference-server` (checked via the GitHub API). When we cannot verify (offline / rate-limited), warn but continue rather than block — an offline user may already have a cached artifact.
3. **Corrupt-tarball recovery.** Validate a cached `.tar.gz` before reusing it. A truncated/interrupted download is now detected, removed, and re-downloaded instead of failing deep in extraction with an opaque `EOFError` traceback.
4. **Download & failure UX.** Use the shared progress bar for release downloads; replace raw extraction tracebacks with an actionable message (delete the bad tarball, re-run).
5. **Ergonomics.** Add a shorter `--reconfig-inf` alias for the flag (long form retained and still shown in `--help`).

## Relationship to #793

Related to #793. Scoping the reconfigure flow addresses the common failure mode described there — the interactive prompt clobbering hand-edited `.env` values — because the flow no longer re-runs the broad env sweep. The `--pull-branch`-specific ask in #793 (that `--pull-branch` alone should force a re-download without the reconfigure flow) is tracked in that issue.

anirudTT:


jashansinghTT (approved):
I was able to change the inference server version in this case and it worked well.

PR #1086, opened 2026-07-17
refactor: consolidate duplicated chipStatus types into shared interface

### LINKED ISSUE #994: Consolidate duplicated frontend chipStatus type into a shared ChipStatus interface
## Summary

The `chipStatus` object shape (`board_type`, `total_slots`, `slots[]`) is duplicated as an inline anonymous TypeScript type across ~8 frontend components, with no canonical exported `ChipStatus` type. The copies have already drifted, and this drift caused a real build break (see context below).

## Why this matters

This is a recurring bug class: a field present at the construction site but absent in a hand-copied receiver type compiles fine until the two are wired together. It just bit us in PR #835:

- `SelectionSteps.tsx` constructs `chipStatus` **with** `board_type` and passes it into `<FirstStepForm chipStatus={chipStatus} />`.
- `FirstStepForm.tsx`'s inline `chipStatus` prop type **omitted** `board_type`, while `renderModelItem` read `chipStatus?.board_type`.
- Result: `error TS2339: Property 'board_type' does not exist ...` — only surfaced at the frontend `tsc` build step, not in `docker compose config` validation.

Fixed for that instance in commit `66f4d44`, but the underlying duplication pattern remains and will recur on the next field addition.

## Evidence of drift

The inline copies are not identical — e.g. `SelectionSteps`'s slot type carries `deployment_id?: number` and `is_multi_chip?: boolean`, while `FirstStepForm`'s slot type omits both.

Components carrying an inline `chipStatus`/slot shape (grep for `total_slots`):
- `app/frontend/src/components/SelectionSteps.tsx`
- `app/frontend/src/components/FirstStepForm.tsx`
- `app/frontend/src/components/DeployModelStep.tsx`
- `app/frontend/src/components/ChipConfigStep.tsx`
- `app/frontend/src/components/MultiCardResetDialog.tsx`
- `app/frontend/src/components/models/ModelsDeployedCard.tsx`
- `app/frontend/src/components/models/RegisterModelDialog.tsx`
- (plus the Voice Agent solution step)

Partial shared fragments already exist but no top-level type:
- `DeviceSlotLike` in `app/frontend/src/utils/deviceFit.ts`
- `ChipStatusSlotLike` in `app/frontend/src/utils/p300x2Placement.ts`

## Proposed fix

Introduce a single exported `ChipStatus` interface (e.g. in `app/frontend/src/utils/deviceFit.ts` or a dedicated `types/` module), reusing `DeviceSlotLike` for `slots`, and have all consumers reference it instead of redeclaring inline object types.

```ts
export interface ChipStatus {
  board_type: string;
  total_slots: number;
  slots: DeviceSlotLike[];
}
```

## Scope / priority

Non-urgent cleanup. Deliberately kept out of the PR #835 build-fix to keep that change minimal. Good candidate for a standalone refactor PR.

AnkittChauhan:
## Summary
Previously, the `chipStatus` object shape (`board_type`, `total_slots`, `slots[]`) was duplicated as an inline anonymous TypeScript type across ~8 different frontend components. Over time, these copies drifted, which recently led to a build break when a field was omitted in a component's local type definition but was still accessed during rendering. 

This change cleans up the technical debt by introducing a canonical `ChipStatus` interface and migrating all consumers to use it, ensuring a single source of truth and preventing future type drift.
- Resolves #994.
## Changes Made
* **Added Shared Type**: Introduced a dedicated module `app/frontend/src/types/chipStatus.ts` which exports the canonical `ChipStatus`, `ChipStatusSlot`, and `ChipSlotStatus` interfaces.
* **Refactored Components**: Removed error-prone inline type declarations from the following components and migrated them to use the new shared `ChipStatus` interface:
  * `SelectionSteps.tsx`
  * `FirstStepForm.tsx`
  * `DeployModelStep.tsx`
  * `ChipConfigStep.tsx`
  * `ModelsDeployedCard.tsx`
  * `RegisterModelForm.tsx`
* **Adopted Partial Types**: Components that fetch or require only a subset of the chip status fields now safely use TypeScript's utility types (e.g., `Pick<ChipStatus, "total_slots">` in `MultiCardResetDialog.tsx` and `VoiceAgentSolutionStep.tsx`).
* **Cleaned up Utilities**: Updated placement logic types (`DeviceSlotLike` in `deviceFit.ts` and `ChipStatusSlotLike` in `p300x2Placement.ts`) to be directly derived from the core `ChipStatusSlot` using `Pick`. This keeps them perfectly in sync with the canonical shape.

## Validation
- `npm run type-check` ✅
- `npm run build` ✅
- Changed-file ESLint: 0 errors ✅


jashansinghTT:
This works and nothing broke. Good addition.

jashansinghTT (approved):
Thanks for the additions, this works well!

PR #1140, opened 2026-07-20
fix: make get_env_var prioritize os.environ over .env

### LINKED ISSUE #804: `get_env_var` in run.py ignores `os.environ`; only reads `app/.env`
## Summary

`get_env_var(var_name, default="")` in `run.py` is named like a generic environment-variable accessor, but it only reads from `app/.env`. It never consults `os.environ`. As a result, shell exports such as `export HOST_PERSISTENT_STORAGE_VOLUME=/somewhere/else` are silently ignored when running `python3 run.py`, which is contrary to what the function name and signature suggest.

This is not breaking any current functionality — every existing call site is internally consistent — but it is a footgun for anyone scripting around `run.py`, writing tests, or running the CLI in an environment where shell-provided env vars are expected to take precedence (CI, containers, ad-hoc overrides).

## The function

`run.py:884`:

```python
def get_env_var(var_name, default=""):
    """Safely get a variable from the .env file."""
    if not os.path.exists(ENV_FILE_PATH):
        return default
    with open(ENV_FILE_PATH, 'r') as f:
        for line in f:
            line = line.strip()
            if line.startswith(f"{var_name}="):
                value = line.split('=', 1)[1]
                return value.strip('"\'')
    return default
```

The docstring is accurate (`from the .env file`), but the function name is not. Most Python code with a name like `get_env_var` follows the convention `os.environ.get(name) or <fallback>`, so readers and callers tend to assume that behavior.

## Why it's a problem in practice

1. **Shell overrides do nothing.** Running `HOST_PERSISTENT_STORAGE_VOLUME=/tmp/foo python3 run.py --cleanup` will not use `/tmp/foo`; the value is whatever is in `app/.env`, or the hard-coded fallback (`TT_STUDIO_ROOT/tt_studio_persistent_volume`). This is the opposite of how `os.environ`-aware tools behave.
2. **Tests need function-level mocking.** Any test that wants to point `run.py` at a temp volume must monkey-patch `run.get_env_var` directly. Setting `os.environ[...]` (the natural first attempt) has no effect, which costs debugging time. I hit this writing E2E coverage for the recent `--cleanup` change.
3. **Wrapper scripts and CI.** A wrapper that exports env vars before invoking `run.py` will appear to work (no error, no warning) while actually doing nothing. The failure mode is silent — the script proceeds with the `.env` value or the default.
4. **`.env`-less environments.** If `app/.env` doesn't exist (fresh clone, container without the file mounted), every call returns the `default`. Real shell env vars that *would* answer the same question are not consulted as a fallback.

## Scope of impact

`get_env_var` is called from **35 sites** in `run.py`. A non-exhaustive sample of the variables it reads:

- `HOST_PERSISTENT_STORAGE_VOLUME` (`run.py:1810`) — used during `--cleanup` to locate `user_config.json`. The case I hit.
- `JWT_SECRET`, `DJANGO_SECRET_KEY`, `DOCKER_CONTROL_JWT_SECRET`
- `HF_TOKEN`, `TTS_API_KEY`, `TAVILY_API_KEY`
- `VITE_APP_TITLE`, `VITE_ENABLE_DEPLOYED`, `VITE_ENABLE_RAG_ADMIN`, `RAG_ADMIN_PASSWORD`
- `DOCKER_CONTROL_SERVICE_URL`, `FRONTEND_HOST`, `FRONTEND_PORT`, `FRONTEND_TIMEOUT`

Every one of these is read in a way that ignores the shell environment.

## Suggested fix

Two reasonable options; I don't have a strong preference, but flagging the trade-offs:

**Option A — make the function behave like its name (preferred).** Consult `os.environ` first, fall back to `app/.env`, then to `default`. Standard layering. Most invasive change is zero (callers don't need to know), and it makes the function match developer intuition.

```python
def get_env_var(var_name, default=""):
    """Get a variable from os.environ, falling back to app/.env, then default."""
    if var_name in os.environ:
        return os.environ[var_name]
    if not os.path.exists(ENV_FILE_PATH):
        return default
    with open(ENV_FILE_PATH, 'r') as f:
        for line in f:
            line = line.strip()
            if line.startswith(f"{var_name}="):
                return line.split('=', 1)[1].strip('"\'')
    return default
```

Risk: a stray env var in the user's shell could now override their `.env`. In practice that's the same trade-off every twelve-factor app makes, and the names involved (`HF_TOKEN`, `JWT_SECRET`, etc.) are already conventional.

**Option B — rename for honesty.** Leave behavior as-is, but rename to `get_env_file_var` (or `read_dotenv`) so the name describes what it does. Lower risk, zero behavior change, but doesn't fix the underlying ergonomics — wrapper scripts still can't override anything.

## Files

- `run.py` — definition at line 884; 35 call sites elsewhere in the file.
- `app/.env` — the file actually being read. No change needed.

## Repro

```bash
# In a checkout where app/.env has HOST_PERSISTENT_STORAGE_VOLUME=/path/from/env
HOST_PERSISTENT_STORAGE_VOLUME=/tmp/should-win python3 -c "
import sys; sys.path.insert(0, '.')
import run
print('get_env_var sees:', run.get_env_var('HOST_PERSISTENT_STORAGE_VOLUME'))
print('os.environ has: ', __import__('os').environ['HOST_PERSISTENT_STORAGE_VOLUME'])
"
```

Expected if the function honored shell env: both lines print `/tmp/should-win`. Actual: `get_env_var` returns the `.env` value; `os.environ` returns `/tmp/should-win`. The shell export is dropped on the floor.

AnkittChauhan:
### Summary
This PR implements **Option A** from the issue description. The `get_env_var` function now behaves intuitively by checking system environment variables before falling back to the local `app/.env` file. 

This enables shell overrides (e.g., `HOST_PERSISTENT_STORAGE_VOLUME=/tmp/foo`), wrapper scripts, and CI/container environments to successfully inject environment variables without them being silently ignored.

- Fixes #804 

### Changes Made
- **Fixed `get_env_var` (`tt_setup/env_config/_dotenv.py`)**: Updated the lookup order to strictly enforce `os.environ` → `app/.env` → `default`.
- **Cleaned up Workarounds**: Removed redundant `or os.getenv(...)` fallbacks in `_metadata.py` and `_fastapi.py` that developers had added to bypass the bug.
- **Added Tests (`tests/test_env_config.py`)**: Added `TestOsEnvironPrecedence` with 4 new tests to guarantee that shell variables correctly override `.env` values, explicitly empty shell variables are respected, and missing shell variables properly fall back to the `.env` file.

### Testing
- ✅ All 22 tests in `test_env_config.py` pass.
- ✅ Verified manually that setting a variable via shell export properly overrides the hardcoded `.env` value at runtime.

### DIFF
--- a/tests/test_env_config.py
+++ b/tests/test_env_config.py
@@ -78,6 +78,48 @@ class TestEnvFileRoundTrip(unittest.TestCase):
         self.assertEqual(existing.get("B"), "2")
 
 
+class TestOsEnvironPrecedence(unittest.TestCase):
+    """os.environ must take precedence over .env file values (issue #804)."""
+
+    def setUp(self):
+        self.tmp = tempfile.NamedTemporaryFile("w", suffix=".env", delete=False)
+        self.tmp.close()
+        self.p = patch.object(_ecfg_dotenv, "ENV_FILE_PATH", self.tmp.name)
+        self.p.start()
+
+    def tearDown(self):
+        self.p.stop()
+        os.unlink(self.tmp.name)
+
+    def test_os_environ_overrides_env_file(self):
+        """Shell export must win over the .env file value."""
+        M.write_env_var("MY_VAR", "from-dotenv")
+        with patch.dict(os.environ, {"MY_VAR": "from-shell"}):
+            self.assertEqual(M.get_env_var("MY_VAR"), "from-shell")
+
+    def test_os_environ_overrides_when_no_env_file(self):
+        """Shell export must work even if .env file does not exist."""
+        os.unlink(self.tmp.name)
+        # Re-create so tearDown doesn't fail
+        open(self.tmp.name, "w").close()
+        with patch.object(_ecfg_dotenv, "ENV_FILE_PATH", "/nonexistent/.env"):
+            with patch.dict(os.environ, {"MY_VAR": "from-shell"}):
+                self.assertEqual(M.get_env_var("MY_VAR"), "from-shell")
+
+    def test_env_file_used_when_no_os_environ(self):
+        """Without a shell override, the .env file value is still returned."""
+        M.write_env_var("MY_VAR", "from-dotenv")
+        with patch.dict(os.environ, {}, clear=False):
+            os.environ.pop("MY_VAR", None)
+            self.assertEqual(M.get_env_var("MY_VAR"), "from-dotenv")
+
+    def test_empty_os_environ_value_still_wins(self):
+        """An explicitly empty shell export must override the .env value."""
+        M.write_env_var("MY_VAR", "from-dotenv")
+        with patch.dict(os.environ, {"MY_VAR": ""}):
+            self.assertEqual(M.get_env_var("MY_VAR"), "")
+
+
 class TestConsistentQuoting(unittest.TestCase):
     """write_env_var must produce ONE consistent (unquoted) format."""
 
--- a/tt_setup/env_config/_dotenv.py
+++ b/tt_setup/env_config/_dotenv.py
@@ -40,7 +40,15 @@ def comment_out_env_var(var_name):
 
 
 def get_env_var(var_name, default=""):
-    """Safely get a variable from app/.env (quotes handled by python-dotenv)."""
+    """Get a variable from os.environ, falling back to app/.env, then default.
+
+    Lookup order (first wins):
+      1. os.environ  – shell exports, CI variables, container env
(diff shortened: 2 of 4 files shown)


jashansinghTT (approved):
This works as intended, maybe in the future we can avoid the issue numbers in the code. Otherwise the change look good. Thanks again for contributing, this is helpful for the team.

PR #1146, opened 2026-07-21
Hotfix: keep Welcome setup completed across restarts

### LINKED ISSUE #1145: run.py --stop should not re-trigger the setup guide unless env vars are missing
## Summary

A plain `python run.py --stop` unconditionally clears the `SETUP_COMPLETE` / Welcome flag. As a result, the next bring-up re-runs the first-run setup guide even though the environment is already fully configured. Stopping the stack should not "un-onboard" the user.

Desired behavior: the setup guide should only reappear when it is actually needed — i.e. when required env variables are missing/incomplete — not simply because the user ran `--stop`.

## Current behavior

In `tt_setup/cleanup/_orchestrate.py` (`cleanup_resources`, the non-`--purge-all` path, ~lines 141–171), stopping deliberately resets the Welcome flag:

```python
# Unset the Welcome flag so the next bring-up re-runs first-run setup,
# while preserving saved secrets (HF token, etc.).
...
with step("Resetting Welcome flag", spinner=False) as s:
    detail = _reset_welcome_flag(...)
```

Running `python run.py --stop` prints:

```
✓ Resetting Welcome flag   SETUP_COMPLETE cleared (sudo)
```

So a normal stop (which is meant to keep persistent data) forces the full setup/welcome flow again on the next start.

## Expected behavior

- `python run.py --stop` (and other non-destructive run modes) should **preserve** the `SETUP_COMPLETE` / Welcome flag and **not** re-trigger the first-run setup guide.
- The setup guide should be shown again **only when required env variables are missing or incomplete** — regardless of which command triggered the run.
- `--purge-all` (full teardown) should continue to reset onboarding state as it does today.

## Suggested approach

- Make the guide's re-appearance driven by an env-completeness check (missing/placeholder required vars) rather than by the stop action clearing the flag.
- Stop clearing `SETUP_COMPLETE` in the plain `--stop` path; leave the flag intact so a subsequent `python run.py` goes straight to bring-up when the env is valid.
- Keep the flag reset behavior only under `--purge-all`.

## Notes
Relevant code: `tt_setup/cleanup/_orchestrate.py` (`cleanup_resources`, `_reset_welcome_flag`, `_sudo_strip_setup_complete`), first-run detection in `tt_setup/env_config/_preferences.py`, and the guide/welcome display in `tt_setup/env_config/_configure.py` / `tt_setup/shell.py`.

jashansinghTT:
Fixes #1145

## Problem

The first-run Welcome wizard (terms, HF token / TTS / Tavily secrets, HF access check) re-appears in the web UI after every stop/start cycle, in both dev mode and production mode. Once a user completes setup, it should stay completed — the guide should only come back when required configuration is actually missing.

## Cause

Two things drove the spurious re-onboarding:

1. A plain `python run.py --stop` (and the deprecated `--cleanup` alias) intentionally stripped the `SETUP_COMPLETE` line from `tt_studio_persistent_volume/backend_volume/user_config.env`. Since the CLI's own Ctrl+C and error hints recommend `--stop`, every normal restart re-triggered first-run setup.
2. The wizard's appearance was driven purely by that flag, so losing `user_config.env` for any incidental reason re-onboarded users whose required secrets were already configured.

## Fix (hotfix)

- `--stop` now leaves `user_config.env` untouched — saved secrets and the Welcome/setup state are both preserved. Removed the now-unused `_reset_welcome_flag` / `_atomic_write_0600` / `_sudo_strip_setup_complete` helpers and updated the `--stop` "Preserved" panel to mention the saved settings.
- `is_setup_complete()` is now env-completeness driven: setup counts as complete when the wizard was finished **or** when the required secret (the HF token) is already configured via `user_config.env` or the environment. The guide only reappears when something it collects is actually missing. The flag alone still counts because the wizard allows skipping every secret — requiring the token on top of the flag would bounce those users straight back to `/welcome`.
- First-run setup still resets on `python run.py --purge-all` (or `--cleanup-all`), which removes the persistent volume and `.env`.

## Verification

- CLI test suite: 163 tests green, including a new regression test asserting a plain `--stop` leaves `user_config.env` byte-identical (`SETUP_COMPLETE` + secrets).
- Backend suite: `shared_config/test_user_config.py` — 21 tests green, including new cases for the completeness check (flag only, token via file, token via env, neither).
- End-to-end on a live stack: marked setup complete via the settings API, ran `python run.py --stop` — the flag survived; restarted — `GET /up/` 200, frontend serving, inference server `{"status":"ok"}`, and `GET /settings/` reports `setup_complete: true`. Separately verified the completeness path: with `user_config.env` absent but `HF_TOKEN` set in the environment, `GET /settings/` still reports `setup_complete: true`, so the app no longer redirects to `/welcome`.

### DIFF
--- a/app/backend/shared_config/test_user_config.py
+++ b/app/backend/shared_config/test_user_config.py
@@ -71,12 +71,30 @@ class TestSaveLoadRoundTrip:
         user_config.save_user_config({"hf_token": "hf_abc123", "evil_key": "x"})
         assert "evil_key" not in user_config.load_user_config()
 
-    def test_setup_complete_round_trip(self, volume):
+    def test_setup_complete_round_trip(self, volume, monkeypatch):
+        monkeypatch.delenv("HF_TOKEN", raising=False)
         assert not user_config.is_setup_complete()
         user_config.mark_setup_complete()
         assert user_config.is_setup_complete()
         content = (volume / "backend_volume" / "user_config.env").read_text()
         assert "SETUP_COMPLETE=true" in content
+
+    def test_setup_complete_when_required_secret_configured(self, volume, monkeypatch):
+        """The Welcome guide is driven by env completeness (issue #1145): a
+        configured HF token counts as complete even without the flag, and a
+        finished wizard counts even if every secret was skipped."""
+        monkeypatch.delenv("HF_TOKEN", raising=False)
+        user_config.save_user_config({"hf_token": "hf_abc123"})
+        assert user_config.is_setup_complete()
+
+    def test_setup_complete_honors_env_var_token(self, volume, monkeypatch):
+        monkeypatch.setenv("HF_TOKEN", "hf_from_env")
+        assert user_config.is_setup_complete()
+
+    def test_setup_incomplete_without_flag_or_token(self, volume, monkeypatch):
+        monkeypatch.delenv("HF_TOKEN", raising=False)
+        user_config.save_user_config({"tavily_api_key": "tvly-x"})
+        assert not user_config.is_setup_complete()
 
 
 class TestValueSanitization:
--- a/app/backend/shared_config/user_config.py
+++ b/app/backend/shared_config/user_config.py
@@ -201,7 +201,19 @@ def get_artifact_info() -> dict:
 
 
 def is_setup_complete() -> bool:
-    return bool(load_user_config().get("setup_complete"))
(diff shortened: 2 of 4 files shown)


anirudTT:
Can you look into this issue and if this PR covers all of it #1145

anirudTT (approved):

PR #1155, opened 2026-07-22
Keep backend runtime state out of the bind-mounted source tree

### LINKED ISSUE #1154: `--purge-all` leaves root-owned files behind in `app/backend/`
`python run.py --purge-all` is supposed to reset the repo to a fresh-clone state, but two things survive it because the backend container writes them into the bind-mounted source tree (`./backend:/backend` in `app/docker-compose.yml`) as root:

1. **`app/backend/db.sqlite3`** — Django's SQLite database, recreated by `migrate --noinput` on every boot (`app/backend/api/settings.py` points `DATABASES` at `BASE_DIR / "db.sqlite3"`). Stale ORM state from the previous install carries over into what should be a clean environment.

2. **`app/backend/**/__pycache__/`** — bytecode compiled by the root container into the host source tree. Stale caches from a previous version survive the reset, and since both are `root:root`, they can't be removed without sudo anyway.

Neither path is in the purge inventory in `tt_setup/cleanup/_orchestrate.py`. Fix would be to add them to the host-state items list — the existing `_remove_path` docker fallback already handles root-owned files.

stisiTT:
Fixes #1154.

## Problem

`./backend` is bind-mounted into the backend container in every mode, and the container runs as root. Two kinds of runtime state therefore land root-owned inside the repo and survive `python run.py --purge-all`:

- `app/backend/db.sqlite3` — recreated by `migrate --noinput` on every boot, so stale ORM state carries over into what should be a fresh environment
- `app/backend/**/__pycache__/` — stale bytecode from previous versions

## Changes

- **Move the SQLite DB into the persistent volume** (`backend_volume/db.sqlite3`). The volume is already mounted in the container, created at import time by `backend_config`, and removed by `--purge-all`, so the DB now resets with everything else and never touches the source tree.
- **Set `PYTHONDONTWRITEBYTECODE=1` for the backend and agent containers.** Set via compose `environment` so it also reaches the prebuilt dev images (dev mode bind-mounts `./backend` and `./agent`), plus the backend Dockerfile for non-compose runs.
- **Add the legacy leftovers to the `--purge-all` inventory** — `app/backend/db.sqlite3`, `app/backend/temp` (RAG upload scratch), and `__pycache__` dirs under `app/backend`/`app/agent` — so existing installs get cleaned up too. The existing `_remove_path` container fallback already handles the root ownership.

## Verification

Ran `python run.py --dev` from a fully purged state:

- backend `/up/` → 200, frontend `:3000` → 200, inference server `/health` → 200
- `db.sqlite3` was created in `tt_studio_persistent_volume/backend_volume/` (not in `app/backend/`)
- `PYTHONDONTWRITEBYTECODE=1` confirmed set inside the running dev container; no new `.pyc`/`__pycache__` written during boot
- `python run.py --purge-all` (aborted at the confirm prompt) lists the new entries: the legacy DB row plus 11 `__pycache__` rows from the previous install


anirudTT:
reviewing this shortly

anirudTT (approved):

PR #1170, opened 2026-07-24
fix: footer shows all device IDs for multi-chip models

### LINKED ISSUE #1147: Footer shows wrong device IDs for multi-chip models (e.g. Llama on device 0,1)
## Summary

The deployed-models popover in the **Footer** shows the wrong device ID(s) for multi-chip models. When a Llama model is deployed across two chips (e.g. `--device-id 0,1`), the footer badge shows only a single device (`dev:0`) instead of both chips (`dev:0,1`).

## Root cause

`app/frontend/src/components/Footer.tsx` (~lines 627–632) renders the device badge from the scalar `model.device_id` only, and ignores the `device_ids` array:

```tsx
{/* Device ID badge */}
{model.device_id != null && (
  <span className="...">
    dev:{model.device_id}
  </span>
)}
```

For a multi-chip deployment the backend reports the full set in `device_ids` (e.g. `[0, 1]`); `device_id` is a single/legacy scalar. The footer only reads the scalar, so it drops the extra chip(s).

`ModelsDeployedCard` already handles this correctly via its `deviceIdsForRow` helper (`app/frontend/src/components/models/ModelsDeployedCard.tsx:55-60`), which prefers the `device_ids` array and falls back to the scalar:

```ts
if (Array.isArray(row.device_ids) && row.device_ids.length > 0) return row.device_ids;
if (row.device_id != null) return [row.device_id];
```

The Footer should use the same logic.

## Steps to reproduce

1. Deploy a Llama model pinned to two chips: `python run.py run <llama-model> --device-id 0,1`.
2. Open the deployed-models popover in the app footer.
3. The device badge shows `dev:0` (only one chip) instead of `dev:0,1`.

## Expected behavior

The footer badge should display all reserved chips for the model, e.g. `dev:0,1`, matching what the Models Deployed card shows.

## Suggested fix

- In `Footer.tsx`, build the badge from `device_ids` (array) with a `device_id` scalar fallback — reuse/extract the `deviceIdsForRow` logic from `ModelsDeployedCard.tsx` so both surfaces stay consistent.
- Confirm the footer's model type (`app/frontend/src/api/modelsDeployedApis.ts`) exposes `device_ids`.

AnkittChauhan:
### Description

The deployed-models popover in the Footer was previously only rendering the scalar `device_id`, which caused it to only display a single chip (e.g. `dev:0`) when a model was actually allocated across multiple chips (e.g. `--device-id 0,1`). 

This PR resolves the issue and aligns the Footer's logic with the `ModelsDeployedCard` component.

- Fixes #1147 
### Changes Made
* Extracted the `deviceIdsForRow` logic into a shared utility (`app/frontend/src/utils/deviceIds.ts`).
* Updated `Footer.tsx` to utilize this shared utility, correctly prioritizing the `device_ids` array when rendering the badge.
* Cleaned up `ModelsDeployedCard.tsx` to use the new shared import.

### Validation & Screenshot
Since I don't have access to Tenstorrent hardware locally to test a multi-chip deployment, I've added an automated E2E test (`footer-device-ids.spec.ts`). This test mocks the backend response with multiple allocated chips (`device_ids: [0, 1, 3]`) and successfully validates that the UI renders the correct multi-chip string (`dev:0,1,3`).

Screenshot 2026-07-24 at 3 41 38 PM

### DIFF
--- a/app/frontend/src/components/Footer.tsx
+++ b/app/frontend/src/components/Footer.tsx
@@ -20,6 +20,7 @@ import { getBuildInfo } from "../api/githubApi";
 import { HardwareIcon } from "./aiPlaygroundHome/HardwareIcon";
 import { cn } from "../lib/utils";
 import { useFooterVisibility } from "../hooks/useFooterVisibility";
+import { deviceIdsForRow } from "../utils/deviceIds";
 
 interface FooterProps {
   className?: string;
@@ -499,10 +500,10 @@ const Footer: React.FC<FooterProps> = ({ className }) => {
                               onClick={handleRefreshBoardDetection}
                               disabled={refreshing || isInCooldown}
                               className={`p-1 rounded-full border border-transparent transition-colors duration-150 ${refreshing
-                                  ? "opacity-70 cursor-wait"
-                                  : isInCooldown
-                                    ? "opacity-60 cursor-not-allowed"
-                                    : "hover:bg-TT-purple-accent/10 dark:hover:bg-TT-purple-accent/20"
+                                ? "opacity-70 cursor-wait"
+                                : isInCooldown
+                                  ? "opacity-60 cursor-not-allowed"
+                                  : "hover:bg-TT-purple-accent/10 dark:hover:bg-TT-purple-accent/20"
                                 }`}
                               aria-label="Retry board detection"
                             >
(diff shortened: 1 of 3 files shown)


jashansinghTT (approved):
This worked for me

PR #1176, opened 2026-07-27
Launcher: --uninstall, --switch, and stale tt-studio shortcut auto-repair

jashansinghTT:
The `tt-studio` shell shortcut bakes in the repo path it was installed from, so launching from a different or newer checkout kept starting the old one — and there was no way to remove the shortcut or move a checkout to a release candidate from the CLI.

- On a normal startup, if the installed shortcut points at a different checkout, it is re-pointed to the current one with a one-line notice.
- `--uninstall`: runs the `--purge-all` teardown and then removes the shell shortcut. One confirmation covers both; declining leaves everything untouched.
- `--switch REF`: fetches origin and checks out a branch (fast-forwarded) or tag (detached, for RCs), then exits so the next run starts on the new code. Refuses on a dirty working tree.

Verified with the launcher test suite (193 tests passing, including new `tests/test_shortcut.py` and `tests/test_switch.py`), plus manual runs: stale-shortcut repair and uninstall against a scratch rc file, `--switch` dirty-tree refusal (exit 1), and the `--uninstall` decline path. A real branch switch / full purge wasn't run on this machine since a live dev stack with deployed models is using the checkout; those paths are covered by the mocked git-sequence tests.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -52,6 +52,8 @@ python run.py --logs          # stream all container logs (docker compose logs -
 python run.py --info          # re-show the "TT Studio is ready" summary (URLs, mode, hardware)
 python run.py --report-bug    # bundle logs (logs/tt-studio-logs-ttbr-*.zip) + open a GitHub issue
 python run.py --install-shortcut # add a `tt-studio` shell shortcut (~/.zshrc/~/.bashrc)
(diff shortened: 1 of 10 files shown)


rnabeelTT (approved):
lgtm

PR #1180, opened 2026-07-28
build(backend): remove the CUDA dependency chain from the backend image

jashansinghTT:
## Problem

Installing the backend crawls through several GB of `nvidia-*` wheels it never uses. `sentence-transformers` was the cause: it requires torch, and torch's Linux wheel hard-depends on 15 CUDA packages plus triton — 4.6 GB of a 5.5 GB `site-packages`. Nothing in the backend imports torch, and no CUDA device is ever attached; we only mount `/dev/tenstorrent`. Its one consumer was the Chroma embedding function for `all-MiniLM-L6-v2`.

## Change

That call site now uses ChromaDB's `ONNXMiniLM_L6_V2` — same weights, running on `onnxruntime`, which was already a requirement. The model is baked into the image, since ChromaDB otherwise downloads it to a path no volume persists and its downloader takes no lock, so the four uvicorn workers would race to extract it.

`Pillow`, `scipy`, and `scikit-learn` are now declared explicitly. All three were already installed, but only as transitive deps of sentence-transformers: `Pillow` for image resizing in `model_control`, and `scipy`/`scikit-learn` because the Dockerfile installs `openwakeword` with `--no-deps` and it imports both at module scope. Without them the wake word feature breaks at import.

A second commit drops `langchain` and `langchain-chroma`, which nothing imports.

## Measurements

| | before | after | |
|---|---|---|---|
| Cold image build | 3 m 35 s | **1 m 28 s** | 2.5x faster |
| `pip install` layer | 101.0 s | **15.6 s** | 6.5x faster |
| Wheels downloaded | 2.89 GB (157) | **147 MB (125)** | 20x less |
| Image, local mode | 11.7 GB | **4.12 GB** | |
| Image, AI Playground mode | 11.7 GB | **2.05 GB** | |
| `site-packages` | 5.5 GB | **771 MB** | |

Method: `docker build --no-cache` of each branch's real Dockerfile, AI Playground variant so the unchanged tt-smi/Rust step doesn't dominate, alternating old/new runs so network variance hits both. Three runs for the pip layer (spread 0.3 s) and two for the full build (spread 2.4 s). Download volume is from `pip download` into an empty cache, which is network-independent. The new build is already paying its ~3 s model bake in these numbers.

The biggest single wheel removed is `torch` at 502 MB, followed by `nvidia_cublas` at 404 MB and `nvidia_cudnn` at 349 MB. This was measured on one machine with a fast link — since the saving is 2.74 GB of transfer, the gap gets wider on slower connections, where the old build is mostly just waiting on the network.

Local mode adds the same unchanged tt-smi/Rust step to both sides, so it shifts both totals by a constant; see the follow-up below.

## Existing collections still work

Verified rather than assumed: the old build wrote a collection, then the new build queried it — same top-3 documents, identical distances to 5 decimal places. Both paths truncate at 256 tokens, which matters because the 1000-character chunk size sits on that boundary. No re-indexing needed.

## Verification

Built both image variants and ran them against an isolated Chroma stack. Cold boot with `--workers 4` reached `GET /up/` 200 with no tracebacks, ingesting the internal knowledge corpus in 83 s. `import openwakeword.model, scipy, sklearn` and the PIL resize both succeed, and the image embeds with `--network none`, proving the bake works offline. `pip list` and `site-packages` contain no `nvidia`, `torch`, `triton`, `cuda`, `transformers`, or `sentence-transformers` entries. Backend tests go from 30 to 36 passing, with the same 2 pre-existing failures that need a live inference server. New `test_embeddings.py` pins the dimensions, norm, and a golden vector so a future ChromaDB bump can't silently swap the model.

Not verified: `tt-smi` on real hardware. That path is untouched and 4.1.2 runs in the built image.

## Follow-up

The `tt-smi` build step leaves its Rust toolchain (~1.5 GB) in the image. Removing it would get local mode to roughly 2.6 GB, but confirming `tt-smi` still works afterwards needs a machine with a Tenstorrent card, so I kept it out of this PR.

One note: dropping the `hf_cache` volume declaration orphans the existing named volume. `docker volume prune` won't collect it — use `docker volume rm app_hf_cache` (~88 MB).


.env.default
```diff
@@ -93,8 +93,9 @@ VITE_LLAMA_AUTH_TOKEN=
 # RAG (required when VITE_ENABLE_RAG_ADMIN=true)
 # ─────────────────────────────────────────────────────────────────────────────
 RAG_ADMIN_PASSWORD=tt-studio-rag-admin-password
-# SentenceTransformer model used to embed documents and queries. Changing this on an
-# existing Chroma volume requires recreating collections — embedding dimensions must match.
+# Name recorded against each collection for the model used to embed documents and
+# queries. all-MiniLM-L6-v2 is the only supported value — it ships inside the backend
```
anirudTT, line 97: maybe we document how to add other embedding models ?

anirudTT:
Can u record a video of the rag working and put it on the PR.
This could also be used as a short video to start hosting as a learning series

anirudTT:
@rnabeelTT : can you test this PR

.env.default
jashansinghTT: Sorry I missed this message. I'll get to it.

anirudTT (approved):

PR #1181, opened 2026-07-28
Get Forge models deploying and serving on Blackhole

stisiTT:
## Heads-up: this uplifts the inference-server artifact to v0.19.0

Worth reviewing on its own, separately from the Forge work. Every model's config comes from that artifact, so this touches more than Forge — the catalog is regenerated from it (54 → 67 models). Forge happened to need it, but the bump stands alone.

## What this fixes

Forge models couldn't be deployed from TT-Studio at all. Four separate things were in the way:

1. **The catalog never picked up Forge models.** The sync script looked for `model_spec.json`, but the released artifact ships its spec as `release_model_spec.json`, so the sync silently never matched and the catalog went stale.

2. **Falcon crashed during model compilation.** The v0.18.0 config (max_num_seqs 4, 16384 context, no `PREFILL_CHUNK_SIZE`) runs the device out of memory during trace capture, and it surfaces as a bare `ValueError: Error code: 3` with no explanation. v0.19.0 ships the tuned config that works — this is the main reason for the bump.

3. **Forge aborted at device init on Blackhole.** No released image sets `TT_MESH_GRAPH_DESC_PATH`, so tt-metal refuses to start on a CUSTOM cluster. Temporarily pinned to the CI build that carries the fix from tenstorrent/tt-inference-server#4785 — **delete that pin once the fix ships in a release.**

4. **Every inference request came back 401.** The media server compares the `Authorization` header against a static API key and never decodes a JWT, so Forge models rejected everything the backend sent and the UI showed nothing. Media and Forge deploys now send that api key instead.

Forge models also get a "Forge" badge in the deploy picker, so it's clear which path a model runs on.

## Tested

Falcon3-7B-Instruct on a P300x2: deploys, completes warmup, and answers queries through the UI. Warmup takes ~10 minutes — that's real trace compilation, not a hang.

## Known gap

`yolox_nano` (Forge CNN) still won't run on Blackhole. It takes the media deploy path, which doesn't get the image override in item 3.

### DIFF
--- a/.env.default
+++ b/.env.default
@@ -20,7 +20,7 @@ BACKEND_API_HOSTNAME=tt-studio-backend-api
 # in commit b788278 (tenstorrent/tt-inference-server). The source branch was deleted but
 # the SHA is still valid. Re-pin to a release version once that fix ships in a tag.
 # TT_INFERENCE_ARTIFACT_BRANCH=b788278762b0bcc9332ad453d265050bfa94738d
-TT_INFERENCE_ARTIFACT_VERSION=v0.18.0
+TT_INFERENCE_ARTIFACT_VERSION=v0.19.0
 # TT_INFERENCE_ARTIFACT_BRANCH=tt-studio/202605
 
 # ─────────────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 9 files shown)


anirudTT:
@stisiTT : can this PR be closed ?

PR #1187, opened 2026-07-29
feat: recover deployments whose container is still running

### LINKED ISSUE #1186: Deployed model becomes invisible to TT-Studio and its chip slot reads as free, blocking redeploys
### What happens

Deploying `distil-large-v3` looks like it fails. Retrying doesn't help — it keeps failing the same way.

What's actually going on is that the model deployed **fine** the first time. The container is up, healthy, and serving requests on its chip. TT-Studio just lost track of it.

Once that happens:

- The model doesn't show up on the Models Deployed page, so it looks like the deploy never worked.
- It isn't selectable as the speech-to-text model in the Voice Agent.
- Its chip slot is reported as **available** even though the container is still holding that chip and its port.

That last point is what makes retrying fail. The next deploy gets handed the same slot, and collides with the container that's already sitting there. So you're stuck in a loop: it looks like it never deploys, and every retry hits the same wall.

### Why it gets stuck

A deployment record can get marked `dead` — for example if a single container lookup fails while docker-control-service is briefly unavailable. Marking a record `dead` is a one-way door: everything that reads deployments back only looks at records marked `starting` or `running`, and nothing ever re-checks whether a "dead" container is actually alive. So one bad lookup permanently hides a perfectly healthy model.

### How to spot it

Compare what Docker says against what TT-Studio says:

```bash
docker port distil-large-v3                                    # 8000/tcp -> 0.0.0.0:7002
docker inspect -f '{{json .HostConfig.Devices}}' distil-large-v3 # /dev/tenstorrent/2
curl -s localhost:8000/docker/chip-status/                     # slot 2: "available"  <-- wrong
```

If the container owns a chip that chip-status calls available, this is it.

### Second, smaller thing

While digging into this, the Whisper container's logs were full of 500s and pydantic tracebacks every 30 seconds. That's the connection warmer sending an LLM-style request (`model` / `prompt` / `max_tokens`) to `/v1/audio/transcriptions`, which expects an uploaded audio file instead. It's harmless to the model, but it makes a working deploy look broken while you're reading logs, which cost real time during this investigation.

### What would fix it

- Don't let one failed container lookup mark a record dead — recheck before believing it.
- If a record says dead but its container is genuinely running, re-adopt it instead of hiding it forever.
- Don't send text-generation warmup payloads to audio and image endpoints.

### LINKED ISSUE #1218: Model deployments can be reported dead while they are still running
## Summary

TT-Studio's backend can only see Docker through the `docker-control-service` on port 8002. Our status logic treated "I couldn't reach that service" as "the container is gone." So any interruption to one host-side service could mark a healthy, running model as **Died Unexpectedly**, hide it from Models Deployed, and free the chip slot its container was still holding.

Found while investigating a Qwen3-32B deploy that ran fine all night but showed up dead in the morning. Related to #1186, which hit the same class of bug from a different angle.

## What happened

The `docker-control-service` stopped but the model container itself was completely unaffected and kept serving for another 13 hours — but the backend was blind to it for that whole window, and partway through, wrote the record off as dead.

The important detail is that **only one service died.** Everything in Docker (backend, frontend, ChromaDB, agent, and the model itself) stayed up. Two of our six services — the inference server (8001) and docker-control-service (8002) — run directly on the host rather than in containers, because they need the Docker socket and `/dev/tenstorrent`. Those two were plain children of the terminal session that ran `run.py`, with no protection against hangup and nothing to restart them.

## Root causes

**1. Status was inferred from failure, not from evidence.** The health monitor decided a container had died by pattern-matching the *text* of the exception. A transport failure (service down) and a real "container not found" answer were treated as the same thing. They are not: only one of them says anything about the
container.

This also had another failure mode. Python connection errors embed a memory address in their text, and one address out of ~16,500 happened to contain the digits `404`, which our code read as a real HTTP 404. That single coincidence is what killed the record.

**2. Marking a record dead was a one-way door.** Nothing ever re-checked a dead record, and everything that reads deployments back only looks at `starting`/`running`. One bad write hid a healthy model permanently — recoverable only by manually re-registering it.

**3. The two host-side services were unsupervised.** No hangup immunity, no restart, and their logs were wiped on every start — so the evidence of *why* a service died was destroyed by the very restart you do to investigate it.

## Real-world consequences

- **A healthy model looks broken.** It vanishes from Models Deployed and appears as "Died Unexpectedly" in Deployment History.
- **Wasted hardware time.** The natural response is to redeploy a large model that was already up and serving — minutes of load time and accelerator time for nothing.
- **Chip/port collisions.** A dead record frees its chip slot while the container still holds that chip and port, so the next deploy is handed the same slot and collides. This is the failure in #1186.
- **Deploys become impossible, silently.** The backend has no fallback path for Docker operations, so while 8002 is down nothing can be deployed or stopped — but the UI still loads and looks healthy, so it presents as a model problem rather than an outage.
- **Losing an SSH session breaks a running setup.** Disconnecting could kill the control plane under a running model, with no recovery until someone re-ran`run.py`.
- **Incidents were undiagnosable.** Log truncation on restart plus ~16,000 identical error lines per outage meant the useful signal was either gone or buried.

jashansinghTT:
Closes #1186

Deploying `distil-large-v3` looked like it failed, but the model had actually deployed fine — TT-Studio had just lost track of the container. Once a deployment record is marked `dead` nothing ever re-checks it, and everything that reads deployments back only looks at `starting`/`running`. So one failed container lookup permanently hid a healthy model, and its chip slot was reported as available while the container still held that chip and port. Every retry then got handed the same slot and collided.

Three changes:

- `health_monitor` requires two consecutive failed lookups before writing `dead`, so a brief docker-control-service hiccup no longer condemns a running model.
- `get_canonical_deployments` re-adopts a `dead`/`exited` record when its container is genuinely running. Records the user stopped themselves are left alone, so stopping a model still frees its slot.
- The connection warmer no longer posts text-generation payloads (`model`/`prompt`/`max_tokens`) to audio and image endpoints. Whisper's `/v1/audio/transcriptions` wants an uploaded file, so every warmup returned a 500 and put a pydantic traceback in the logs of a perfectly healthy container every 30 seconds.

### Verification

Against the running dev stack on a P300x2 box with Llama-3.1-8B, `distil-large-v3`, and `speecht5_tts` deployed. `/up/`, `/models/health/`, `:8001/health` and the frontend all returned 200.

- Forced the live Whisper record to `dead` and hit `/docker/chip-status/`: slot 2 came back `occupied` on the first request, the record was rewritten to `running`, and the model reappeared in `/docker/deployments/` with `device_id=2`.
- Repeated it with `stopped_by_user=True`: the record stayed `dead` and slot 2 correctly freed, confirming a user stop is still honoured.
- Watched the Whisper container across a full 30s warmup cycle: 500 count held at 16 with no new entries (last one predates the change).

`docker_control/` tests fail to collect and `model_control/test_model_api.py` fails on this box, both identically with these changes stashed — pre-existing, unrelated to this PR.


anirudTT:
@rnabeelTT : noticed this error , will try testing this as well

rnabeelTT:
As per our discussion, we will be combining our efforts in this PR to resolve the original issue as well as #1208

rnabeelTT:
## Root Causes
Documented in detail in #1218 

## Fixes

- Status is only ever changed on a **definitive answer** from a service that actually replied — a reported non-running Docker status, or a real 404. An unreachable service now changes nothing as far as the model container status is concerned, because they can still be interacted with. Genuine crashes are still detected on the next 5s poll, so nothing is reported more slowly.
- A record marked dead is **re-adopted** if its container turns out to be alive, so a stale record heals itself instead of needing manual re-registration (from #1187, with the name-matching narrowed so it can't revive unrelated historical records that share a container name).
- During an outage, Models Deployed **shows the last known state** instead of going blank, and the store is left untouched.
- The two host-side services now **ignore terminal hangup, run in their own session, and restart themselves.** Closing SSH no longer affects them.
- Service logs are **archived per restart** (matching the existing `model_run_logs` convention) instead of being overwritten, and outages log once per transition instead of once per container per poll.

rnabeelTT (approved):
this is complete, needs one more test

rnabeelTT:
@jashansinghTT can you please retest this branch once before merging?

PR #1189, opened 2026-07-30
fix: disable spinners on host service cleanup to keep sudo prompts visible

### LINKED ISSUE #1171: [UX/CLI] python3 run.py --stop hangs silently on sudo password prompt during "Stopping inference API..."
### Description

When executing `python3 run.py --stop`, the CLI displays an animated Rich spinner at step **`Stopping inference API...`**. If host processes or ports (8001/8002) require `sudo` to inspect or terminate, `cleanup_fastapi_server` / `kill_process_on_port` triggers `sudo` under the hood.

Because `sudo` prompts for credentials while the Rich live spinner is actively redrawing the terminal line:
1. No message/warning informs the developer that a `sudo` password is required.
2. The `Password:` prompt is hidden/masked behind the animated spinner (`⠋ Stopping inference API...🔑`).
3. Developers are left waiting indefinitely thinking the process is hung or processing in the background.

---

### Context (Relation to PR #1074)

PR #1074 (*"Remove sudo password prompt from setup"*) successfully removed proactive `sudo` prompts during startup (`python run.py`) by utilizing `_remove_path_with_docker()` (ephemeral Docker container cleanup). However, the `--stop` cleanup path (`python run.py --stop`) was not fully covered: stopping host processes (FastAPI / Docker control on ports 8001/8002) still falls back to interactive `sudo` inside the live spinner context.

---

### Suggested Fixes

1. **Explicit Prompt Notification & Pre-authentication (Immediate UX Fix)**:
   Before triggering `sudo` commands or entering the spinner context, print a clear notification informing the user that root privileges are required:
   ```python
   console.print("[warning]Enter your sudo password:[/warning]")
   subprocess.run(["sudo", "-v"], check=False)
2. **Extend PR #1074's Non-Sudo Docker Approach**: Align the `--stop` flow with the non-sudo philosophy introduced in this PR:
- Ensure host services (FastAPI server / Docker control) run under non-root user permissions where possible to avoid root ownership of ports 8001/8002 altogether.
- For file/process cleanup, utilize Docker-based management (`_remove_path_with_docker()`) or non-interactive signals rather than host `sudo kill`.

### Screenshot/ScreenRecord

https://github.com/user-attachments/assets/fe261c66-0a61-4837-850b-088600956d70

AnkittChauhan:
### Description
During runtime cleanup, host service cleanup (`cleanup_fastapi_server` and `cleanup_docker_control_service`) can discover a `sudo` requirement **after** the initial preflight check—for example, when encountering stale/root-owned PID files or when `lsof` cannot identify the owning process.

Previously, these steps ran inside animated step spinners (`step(..., spinner=True)`). When a late `sudo` prompt occurred during the animated step, the prompt was obscured or hidden by the spinner updates, leading to user input hangs.

- Fixes #1171 

### Changes
- **`tt_setup/cleanup/_runtime.py`**: Set `spinner=False` for "Stopping inference API" and "Stopping Docker control" cleanup steps so late `sudo` prompts remain visible and interactive.
- **`tests/test_cleanup.py`**: Added `test_host_service_cleanup_steps_never_hide_sudo_prompts` to verify these steps execute non-animated.

### Validation & Screenshot
- Added and ran unit test `test_host_service_cleanup_steps_never_hide_sudo_prompts`.

https://github.com/user-attachments/assets/a54b9179-8f73-4b69-b5bf-4aebeaaa01fe

### DIFF
--- a/tests/test_cleanup.py
+++ b/tests/test_cleanup.py
@@ -478,6 +478,34 @@ class CleanupDockerSurfaceTests(unittest.TestCase):
         order = self._record_runtime_order(args)
         self.assertEqual(order[:3], ["deployments", "compose_down", "network_rm"])
 
+    def test_host_service_cleanup_steps_never_hide_sudo_prompts(self):
+        """Host cleanup may discover sudo is needed after the preflight probe.
+
+        Keep both service-cleanup steps non-animated so a late sudo prompt is
+        always visible, even when lsof cannot identify the owning process or a
+        stale PID file points at a root-owned process.
+        """
+        args = SimpleNamespace(dev=False, no_sudo=False, cleanup_all=False)
+        step_calls = []
+
+        @contextlib.contextmanager
+        def fake_step(label, spinner=True):
+            step_calls.append((label, spinner))
+            yield SimpleNamespace()
+
+        with patch.object(_cl_runt, "_docker_daemon_status", return_value="down"), \
+             patch.object(_cl_runt, "_port_owned_by_root", return_value=False), \
+             patch.object(_cl_runt, "step", side_effect=fake_step), \
+             patch.object(_cl_runt, "cleanup_fastapi_server"), \
+             patch.object(_cl_runt, "cleanup_docker_control_service"), \
+             contextlib.redirect_stdout(io.StringIO()):
+            run._cleanup_runtime(args, has_docker_access=True)
+
+        self.assertEqual(step_calls, [
+            ("Stopping inference API", False),
+            ("Stopping Docker control", False),
+        ])
+
 
 class TermsAcceptanceGateTests(unittest.TestCase):
     def test_accepting_terms_persists_prefs_and_gates_off_first_run(self):
--- a/tt_setup/cleanup/_runtime.py
+++ b/tt_setup/cleanup/_runtime.py
@@ -87,7 +87,13 @@ def _cleanup_runtime(args, has_docker_access):
                       "host service (ports 8001/8002).[/warning]")
         subprocess.run(["sudo", "-v"], check=False)
 
-    with step("Stopping inference API", spinner=True):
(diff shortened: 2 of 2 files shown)


jashansinghTT:
Hey @AnkittChauhan ,  can you check whether you were on root user or not? If you're not on a root user, I think then it asks you for sudo, but for us in a new fixes it doesn't ask us for sudo when we do --stop. Let me know once you update your changes according to the current dev branch. If still this remains an issue, then we both can look into it further.

jashansinghTT:
Also what software are you using to record the terminal? It looks really nice.

rnabeelTT:
I think the better fix for this would be to ensure we always use the docker container method to implement the `--stop` functionality. It seems that my PR #1074 may have missed a path. Ideally, we don't want a sudo prompt at all.

AnkittChauhan:
@rnabeelTT : You’re right that this is a gap left by PR #1074, i also mentioned that in in this Issue #1171. That PR handles root-owned files through the cleanup container, but the --stop path still performs host-process cleanup for FastAPI and Docker Control, which can require sudo.

@jashansinghTT : This PR addresses the immediate UX issue by ensuring any late sudo prompt remains visible. Fully eliminating sudo from --stop would require a broader redesign of how these host services are managed, so I suggest treating that as a follow-up unless you prefer expanding this PR.

AnkittChauhan:
> Also what software are you using to record the terminal? It looks really nice.

@jashansinghTT : Yess Fr 😅, I recorded via Mac's in-build screen recorder, when you do `cmd + shift + 5` it's gives you a option to record selected window, select terminal there.

jashansinghTT (approved):
I couldn't replicate the issue on my end but the changes look good and didn't break anything. The refactor makes sense as well hence I would approve this at the present.

PR #1190, opened 2026-07-30
fix: rag upload failures for .log files, drag-drop double uploads, and broken empty collections

jashansinghTT:
## Problem

Uploading a `.log` file to RAG (reported with a tt-studio backend log) fails in a confusing cascade:

1. The document processor only accepted `.pdf/.txt/.docx/.md/.html` and a few code extensions, so the insert returned 400 `Unsupported file type` — but the collection had already been auto-created, leaving an **empty datasource** behind.
2. The empty collection was then mislabeled **Internal Knowledge** by a heuristic ("no documents + no last_uploaded_document metadata"), which disabled its Delete/Upload buttons, so it couldn't be retried or cleaned up from the UI.
3. Chat queries against the empty collection only returned the merged Tenstorrent docs corpus, so the model answered that no log data exists.
4. On top of that, drag-and-drop fired every upload **twice**: the dropzone inside `GentleFileUpload` uploads the dropped files, and the drop event then bubbles to the wrapping `Card`'s own `onDrop`, which uploaded them again — hence the duplicated "Uploading document..." and error toasts.
5. The error toast never included the backend's reason, the in-progress toast used a green success checkmark, and a failed upload left the row spinner stuck.

## Changes

- **Backend**: accept `.log/.json/.csv/.xml/.yaml/.yml` as plain-text uploads in `DocumentProcessor`; `.log` files group under the `text` folder type.
- **Frontend**:
  - `Card.onDrop` no longer re-uploads dropped files (dropzone already handles them); it only resets drag styling.
  - If the document upload fails right after auto-creating its collection, the collection is rolled back so no orphaned empty datasource remains.
  - Only explicitly system-tagged collections (`metadata.type === "internal_knowledge"`, `created_by === "system"`, or the seeded collection name) are treated as Internal Knowledge — empty user collections keep working Delete/Upload buttons.
  - Upload failure toasts now include the backend's error message (e.g. "Unsupported file type"), the uploading spinner clears on failure, and the in-progress toast is informational instead of a success checkmark.
  - File-picker `accept` lists now match what the backend supports: `.log` added, `.doc` removed (legacy Word binaries were never parseable by python-docx).

## Verification

- Exercised the document-processor change inside the `backend-dev` Docker image (same runtime deps as the deployed backend): a `.log` file now processes and chunks correctly, `.json/.csv/.yaml/.yml/.xml` map to the text handler, and unsupported types (e.g. `.zip`) are still rejected with `Unsupported file type`, which now propagates into the UI toast.
- Frontend: `tsc --noEmit` clean, ESLint clean on changed files (only pre-existing `no-console` warnings), production `npm run build` succeeds.
- Full in-browser end-to-end wasn't run: the TT-Studio stack currently running on this machine is served from a different user's checkout, so this branch couldn't be deployed against it without disrupting their session.

## Review follow-ups

Addressed the review feedback:

- **Selected datasources now actually answer from their own documents** (@rnabeelTT's report): single-collection queries merged the shared Tenstorrent docs corpus into results sorted purely by embedding distance, so the 1000+-page internal knowledge corpus could crowd the user's uploaded chunks out of the response entirely — chat then answered from docs and claimed the uploaded data didn't exist. The merge is now primary-first: the selected collection's matches take the result slots first, and internal knowledge only backfills leftover capacity (e.g. for small or empty collections).
- The `Unsupported file type` error now reports just the file extension instead of the server-side temp path, since the message is surfaced in UI toasts (Copilot).
- `.json/.csv/.xml/.yaml/.yml` uploads are now classified into the `text/` virtual folder, consistent with their processing as plain text (Copilot).
- All `collectionsUploading` state updates use functional updates with dedupe, so concurrent uploads can't clobber each other's spinner state (Copilot).

Verified in the `backend-dev` image (no running stack needed): primary-first merge keeps the selected collection's chunks and backfills/falls back to internal knowledge correctly, `.log` files process and chunk, the new extensions map to the text handler, and unsupported types are rejected with a path-free message. Frontend re-verified after the `dev` merge-in: `tsc --noEmit` clean, ESLint clean on changed files (pre-existing `no-console` warnings only), production build succeeds.

### DIFF
--- a/app/backend/vector_db_control/document_processor.py
+++ b/app/backend/vector_db_control/document_processor.py
@@ -27,6 +27,13 @@ class DocumentProcessor:
         '.ts': 'application/typescript',
         '.tsx': 'application/typescript',
         '.jsx': 'application/javascript',
+        # Plain-text formats routed through process_text
+        '.log': 'text/plain',
+        '.json': 'text/plain',
+        '.csv': 'text/plain',
+        '.xml': 'text/plain',
+        '.yaml': 'text/plain',
+        '.yml': 'text/plain',
     }
 
     @staticmethod
--- a/app/backend/vector_db_control/document_processor.py
+++ b/app/backend/vector_db_control/document_processor.py
@@ -134,7 +134,10 @@ class DocumentProcessor:
         """Process document based on file type."""
         file_type = cls.get_file_type(file_path)
         if not file_type:
-            raise ValueError(f"Unsupported file type: {file_path}")
+            # Report only the extension — the full path is a server-side temp
+            # location and this message is surfaced to the user in the UI.
+            extension = os.path.splitext(file_path)[1] or os.path.basename(file_path)
+            raise ValueError(f"Unsupported file type: {extension}")
 
         processors = {
             'application/pdf': cls.process_pdf,
--- a/app/backend/vector_db_control/views.py
+++ b/app/backend/vector_db_control/views.py
@@ -369,7 +369,7 @@ class VectorCollectionsAPIView(ViewSet):
             elif file_extension in ['.doc', '.docx']:
                 folder_type = "docs"
                 folder_path = f"docs/{filename}"
-            elif file_extension in ['.txt']:
+            elif file_extension in ['.txt', '.log']:
                 folder_type = "text"
                 folder_path = f"text/{filename}"
             elif file_extension in ['.ppt', '.pptx']:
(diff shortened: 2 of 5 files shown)


rnabeelTT:
it does let me upload json and log files but it doesnt answer from them. it seems that even if i dont select the internal knowledge collection, it still has access to that.

rnabeelTT (approved):
LGTM

PR #1198, opened 2026-07-31
feat: hybrid RAG retrieval with reranking, structure-aware chunking, and server-side orchestration

jashansinghTT:
## What

- New `POST /collections/retrieve` endpoint runs the whole RAG pipeline server-side: optional LLM query rewrite → dense + BM25 hybrid retrieval → reciprocal-rank fusion → cross-encoder reranking → parent-section expansion → relevance threshold → token budget. The chat UI now makes this one call instead of assembling context in the browser; the old `/query` endpoints are unchanged.
- Reranker is a quantized ONNX ms-marco cross-encoder on the existing onnxruntime — no torch, so it stays compatible with the CUDA-removal work in #1180 (both PRs touch `requirements.txt`; whichever lands second has a trivial merge).
- Ingestion is now two-stage: markdown/HTML split on headings and PDFs per page into parent sections, then small child chunks embedded with contextual headers (`Document: … — Section: …`) and `parent_id`/`parent_text` metadata. Deterministic chunk ids + upsert mean re-uploading a file no longer duplicates its chunks.
- The search agent gets a `document_search` tool over the internal knowledge corpus and a prompt that routes between web and docs (per-user collections need identity plumbing through the agent path — follow-up).
- Everything is env-tunable (`RAG_QUERY_REWRITE`, `RAG_RERANK_*`, `RAG_CHUNK_*`, `RAG_CONTEXT_TOKEN_BUDGET`); rewrite defaults off, rerank defaults on with graceful offline fallback.
- The rerank threshold can no longer empty the results: the best `RAG_RERANK_FLOOR` chunks (default 2) always survive it, and the default `RAG_RERANK_MIN_SCORE` dropped from 0.1 to 0.05 (per review feedback — on niche corpora the cross-encoder can score every candidate low, and a few weak chunks ground the answer better than none).

## Verification

- 36 new unit tests (`test_retrieval.py`, `test_chunking.py`) pass in the backend image.
- Exercised end-to-end against a scratch Chroma + backend: corpus reseed with the new pipeline, upload/re-upload chunk-count stability and stale-chunk sweep, ownership 403s, reranked hybrid results with per-stage signals.
- New eval script (`vector_db_control/eval/`, 30 questions over the internal corpus): full pipeline **83% hit@5 / 57% hit@1** vs dense-only **60% / 40%**.
- Frontend `type-check` clean; new deps attributed in `LICENSE` (attribution gate passes).


anirudTT:
@jashansinghTT : please review the copolit comments and address them

then can you tag @rnabeelTT to help review this PR.

jashansinghTT:
Hii @rnabeelTT , can you review this PR?

rnabeelTT:
This is great progress! I suggest lowering the default RAG_RERANK_MIN_SCORE to ~0.02-0.05 instead of 0.1 or make it a floor of N approach as on my custom dataset i noticed that the reranker would sometimes drop everything and return empty results, which seems worse than having some chunks with less scores.

rnabeelTT (approved):

jashansinghTT:
Good catch @rnabeelTT — done in e65dac1, with both halves of your suggestion:

- `RAG_RERANK_MIN_SCORE` default lowered from 0.1 to 0.05.
- Floor-of-N: the best `RAG_RERANK_FLOOR` reranked chunks (default 2, env-tunable) now always survive the threshold, so retrieval can never come back empty just because the cross-encoder scored a niche corpus low. The threshold still trims low scorers whenever enough candidates pass it.

Covered by four new unit tests on the extracted `apply_rerank_threshold` helper (36 tests passing in the backend image).

Issue #977, opened 2026-06-22
Consolidate scattered model/artifact version pins into one file


anirudTT:
**Goal:** Today a single logical "pin" is edited across `app/.env.default`, `run.py` (`MODEL_SPECS_ENV`), and hardcoded image overrides in `app/backend/docker_control/` — tracked only by code comments. Collect them into one file (e.g. `pins.yaml`) with `reason` / `tracking_issue` / `expires` per entry, plus a CI check that warns when a pin is past its expiry or its tracking issue is closed. `run.py` and the deploy path read from this one file instead of N hand-edited locations.

Lower priority / follow-up.

Part of #960.

AnkittChauhan:
@anirudTT : /assign

stisiTT:
Hit a concrete instance of this while getting an experimental dev-tier-only model (Qwen3.5-9B) deployable: `sync_models_from_inference_server.py` silently drops any hand-curated catalog entry on every resync, with no error.

**Root cause:** `main()` (`sync_models_from_inference_server.py:334-360`) does a full destructive rebuild — `models = normalize(source_path)` then `json.dump(catalog, ...)` straight to `OUTPUT_JSON`, with no read of the existing file first. `normalize()` builds each entry from a fixed key list (lines 300-315) derived purely from the fetched source JSON. Two consequences:
- A field that only ever gets set by hand (e.g. a `requires_dev_catalog` flag) has nowhere to survive a resync, since nothing in the source JSON produces it.
- A model that only exists in tt-inference-server's dev-tier catalog can't appear at all when synced from `release_model_spec.json` (a prod-only release snapshot) — so the entry itself, not just one field on it, disappears every time `--resync`/`--pull-branch`/`--reconfigure-inference-server` runs.

Either way: someone hits a bug, fixes it by hand-editing the catalog, then loses the fix on the next sync with no warning — it just quietly reverts.

**Proposed fix:** make the sync additive instead of destructive. Before writing, load the existing `OUTPUT_JSON` (if present) into a `by_model_name` map, then:
1. For any model name present in both the old and freshly-synced catalog, carry forward fields the sync script itself never derives (currently just `requires_dev_catalog`) from the old entry onto the new one.
2. For any model name present in the *old* catalog but absent from the fresh one (e.g. a dev-tier-only model the source JSON can never mention), keep the old entry as-is rather than dropping it.

[...]

PR #1201, opened 2026-08-06
feat: add per-model dev-catalog override for deploys

stisiTT:
## Summary

Experimental models couldn't be deployed at all. This makes them deployable, with no change to models that already work.

## The problem

The inference server keeps two model catalogs — a stable one and a richer experimental one — and TT-Studio only ever used the stable one, so an experimental model failed instantly with an error that read like a bug. Getting one working surfaced three further blockers behind it.

## What changed

- **A model can opt into the experimental catalog** for its own deploy. Every other model is untouched.
- **Its container image is now supplied automatically.** Experimental entries don't carry one the way stable entries do, and the deploy was rejected without it.
- **Experimental deploys now run in isolation.** They previously shared one long-lived process and could briefly corrupt other people's catalog lookups mid-deploy.
- **A model can name the exact inference-server build it needs, per board.** That used to be a single setting for the whole installation, kept in a file a full reset erases — so it silently reverted and the next deploy failed for a reason that looked unrelated. It now travels with the model and survives a reset.
- **Refreshing the catalog no longer discards hand-set values.** It rebuilt from scratch every time, so a fix applied by hand vanished on the next refresh with no warning. It now merges. (Refs #977.)

## Status

Verified end-to-end on real hardware, including the case this was built for: with the installation deliberately left on a build that does not contain the model, the deploy succeeded against the model's own build, the shared build was untouched, and a catalog refresh preserved the settings. Naming a build that doesn't exist fails immediately and says which one.

### DIFF
--- a/app/backend/docker_control/docker_utils.py
+++ b/app/backend/docker_control/docker_utils.py
@@ -12,7 +12,7 @@ from django.core.cache import caches
 
 from shared_config.device_config import DeviceConfigurations
 from shared_config.logger_config import get_logger
-from shared_config.model_config import model_implmentations, _impl_id
+from shared_config.model_config import model_implmentations, _impl_selector
 from shared_config.backend_config import backend_config
 from shared_config.model_type_config import ModelTypes
 from shared_config.user_config import get_tavily_api_key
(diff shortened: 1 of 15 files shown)


anirudTT (approved):
LGTM, mechanism looks good — just need to fix the PR title to follow Conventional Commits format (e.g. feat: add per-model dev-catalog override for deploys) so the linter passes before merge

PR #1202, opened 2026-08-06
Add tests for dev-catalog subprocess deploys

stisiTT:
## Summary

Adds test coverage for the subprocess-based dev-tier deploy path added in #1201. No existing test exercised that mechanism directly.

## What it does

Builds a minimal fake `run.py` and runs it as a real subprocess (not mocked) to check:
- Output correctly reaches the deploy's progress and log tracking
- The subprocess never touches the parent process's own environment, working directory, or arguments
- A failed deploy retries exactly once with adjusted options, and stops retrying if it fails again

No real hardware or artifact needed to run these.

### DIFF
--- a/inference-api/tests/__init__.py
+++ b/inference-api/tests/__init__.py
@@ -1,0 +1,2 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
--- a/inference-api/tests/test_dev_mode_subprocess.py
+++ b/inference-api/tests/test_dev_mode_subprocess.py
@@ -1,0 +1,267 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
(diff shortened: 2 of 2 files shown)


anirudTT (approved):
Nice coverage. One follow-up: right now nothing in CI actually runs this — `launcher-tests.yml` is the only workflow that runs a Python suite, and it's path-filtered to `tt_setup/**`/`tests/**`/`pyproject.toml`, so it won't pick up `inference-api/tests/`.

Next step could be: add a new workflow (e.g. `inference-api-tests.yml`) triggered on:
```yaml
on:
  pull_request:
    branches: [dev, main, staging]
    paths:
      - "inference-api/**"
      - ".github/workflows/inference-api-tests.yml"
```
Scoping to `inference-api/**` so changes to `api.py` itself also trigger the tests, not just edits to the test file.

stisiTT:
Superseded by #1201. The commit from this branch (`Add tests for dev-catalog subprocess deploys`) is now folded directly into #1201, along with the catalog-sync source detection fix, the structured-progress fix, and the non-chat deploy impl_id fixes. Closing to keep a single reviewable PR.

PR #1207, opened 2026-08-06
feat: add --purge-model to uninstall specific models

jashansinghTT:
Reclaiming space from one unused model currently means `--purge-all`, which wipes every model plus the .env and all persistent data. `--purge-model` removes just the named model(s) — weights, docker volume, env file, and any running container — and leaves everything else alone. Run bare, it opens a numbered picker of installed models with sizes; on the command line it takes one or more names (repeatable flag, comma form, or an unambiguous partial name). Shared docker images are reference-counted against the catalog, so a tag is only removed when no other installed model still uses it.

Same safety UX as `--purge-all`: full inventory with sizes before anything is deleted, a confirmation prompt that `--yes` skips, and all-or-nothing name validation (a typo aborts before anything is touched).

- [x] New `tt_setup/cleanup/_purge_model.py` (discovery, picker, orchestrator) + targeted docker helpers in `_resource_ops.py`
- [x] Bare `--purge-model` picker via an argv sentinel in `main()` (typer's vendored click can't express optional option values)
- [x] Shared `_confirm_purge` extracted from `cleanup_resources` so both purge flows confirm identically
- [x] 31 new tests (name matching incl. the `X` vs `X-FP8` collision, image ref-counting, abort-before-destructive, full removal); suite at 233 green
- [x] Verified end-to-end under a real PTY against a scratch persistent volume: picker → inventory → confirm → weights/env removed, deployment record marked stopped; unknown name exits 1 with suggestions
- [x] Docs: new `dev-docs/purge-model-guide.md`, run-py-guide options table + Common Operations, CLAUDE.md, `--help-env` examples


stisiTT:
Took a careful pass through this. The destructive side is solid — the name matching holds up against every model in the catalog, including the lookalike pairs that would be easy to get wrong, and nothing is touched before the confirmation prompt. I don't think this deletes the wrong thing.

Three things I'd want addressed before it ships. All of them are about what the command *reports*, not what it does:

**1. With Docker stopped, it says the model's storage wasn't there — rather than that it couldn't check.** That's where most model weights actually live, so someone can run this, be told it completed, and still have tens of gigabytes sitting on disk. Given the whole point is reclaiming space you can trust is gone, this is the one I'd fix first.

**2. When something still running blocks removal, the suggested recovery step doesn't help.** The command it points people to deliberately leaves running models alone, so following the advice changes nothing and they hit the identical error on the retry.

**3. If deleting the files fails, the run still reports success** — including a reclaimed-space figure that was calculated before anything was deleted. The underlying warning gets captured into the log instead of shown on screen. Most likely to come up when file ownership needs elevated permissions.

All three look like small, contained changes. Worth adding coverage for those three paths as well — they're the failure modes the current tests don't exercise, which is why they slipped through.

Happy to walk through the line-level detail if that's useful, and I have a few smaller non-blocking notes I can pass along separately.

jashansinghTT:
Thanks @stisiTT , all of these are valid suggestions and I will work on these

rnabeelTT (approved):
LGTM

PR #1212, opened 2026-08-07
feat: pin Voice Agent Solution LLM to Qwen3.5-9B on Blackhole

stisiTT:
## Summary

Swaps the Voice Agent Solution's LLM to Qwen3.5-9B on Blackhole hardware, building on the Qwen3.5-9B catalog entry from #1210. Where Qwen3.5-9B is compatible the wizard pins it and shows a fixed LLM card; where it is not, the wizard falls back to Qwen3-8B, and only then to the Instruct-preferring chat dropdown, so the pipeline still deploys.

## What changed

- **Pin Qwen3.5-9B** as the pipeline LLM when the board reports it compatible, rendered as a fixed card (no dropdown) and flagged experimental.
- **Fall back to Qwen3-8B** where Qwen3.5-9B is not compatible. Qwen3-8B was not in the catalog, so it is hand-added from the v0.19.0 artifact spec and marked `hand_owned` — `merge_hand_owned` displaces the same-named source entry on a future resync rather than duplicating it.
- **Device placement.** Qwen3.5-9B is single-chip, so the pinned pipeline is one device per stage (0/1/2, 3-slot pre-flight). Qwen3-8B and Llama-3.1-8B run across a whole P300 card, so on a P300x2 the fallback LLM takes slots 0,1 with Whisper on 2 and SpeechT5 on 3 (4-slot pre-flight).

Note on scope: the pinned path is the common case on Blackhole, since Qwen3.5-9B is compatible on every Blackhole board. The 0,1 pairing therefore only engages on a P300x2 where Qwen3.5-9B is unavailable — it is a safety net, and it restores the card-pair placement Llama-3.1-8B had before this branch.

## Verified

- Qwen3-8B loads into the model registry as `CHAT` with `chips_required: 1`, all six device configurations resolving, and `is_compatible: true` on P300x2.
- The Qwen3-8B image manifest exists on GHCR.
- Catalog JSON is valid with no duplicate `model_name`.
- The Qwen3-8B name matcher was checked against every catalog entry — it matches only Qwen3-8B and the Llama-3.1-8B rows, not Qwen3-32B, Qwen3-4B, Qwen3.5-9B or Qwen3-Embedding-8B.
- Earlier commit confirmed against the live backend on a P300x2 box: Qwen3.5-9B + Whisper + SpeechT5 all report compatible and the pinning logic selects Qwen3.5-9B.

## Not yet verified

- [ ] Full Voice Agent Solution end to end on hardware (LLM + Whisper + TTS serving)
- [ ] A Qwen3-8B deploy on the fallback path
- [ ] `distil-large-v3` and `speecht5_tts` deploys returning 201 rather than 422 (the open items from the `impl` string fix in #1201)

## Dependency reviewers should know about

The Qwen3.5-9B deploy fixes from #1217 are **not on `dev`**. That PR was closed on the understanding they had landed in its base branch, but #1210 squash-merged as the catalog entry alone. Missing: the switch that disables the 5s metal operation timeout (first-load weight remaps abort without it), the `.artifacts/refs` mount and workflow-log discovery for ref-based deploys (logs 404 in the UI), and the `qwen3_coder` tool-call parser for this Qwen family.

Pinning the voice pipeline to Qwen3.5-9B is unlikely to work end to end until those land, so this should probably merge after a fix branch that recovers them.


anirudTT (approved):
Tested this, works.

One question on scope: do we want to fully commit the Voice Agent's LLM to Qwen3.5-9B, or keep it as the default while still letting users deploy Llama-3.1-8B?

Follow-up task: the thinking/reasoning trace defaulting to open needs to be fixed for Qwen3.5-9B.

jashansinghTT (approved):
Verified on a P300x2 (dev mode): Qwen3.5-9B + distil-large-v3 + speecht5_tts deploy to slots 0/1/2 and the pinned LLM card renders as described. Pushed two commits to the branch: a merge of dev that resolves the catalog JSON conflict (kept dev's Qwen3.8-27B artifact ref alongside the new Qwen3-8B entry), and a voice-agent fix that sends chat_template_kwargs.enable_thinking=false so Qwen3.5-9B answers in ~5s instead of streaming 30s+ of reasoning into the transcript and TTS.

jashansinghTT (approved):
LGTM

PR #1217, opened 2026-08-10
feat: fix qwen3.5-9b deploy timeouts and workflow log streaming

jashansinghTT:
## Summary

Follow-ups to #1210 (Qwen3.5-9B on Blackhole) that dev does not have yet. This PR was previously based on the pre-rewrite `stisi/feat-qwen35-9b-p150` branch; it has been rebuilt on `dev`, dropping everything #1210 already landed (the catalog entry, published image pin, and artifact-ref deploy path). What remains:

- **Metal op timeout** — first-load weight remaps and kernel compiles on experimental Blackhole builds can exceed the default 5s `TT_METAL_OPERATION_TIMEOUT_SECONDS` and abort a healthy deploy. `DeployView` now passes `disable_metal_timeout` through the inference API to `run.py --disable-metal-timeout`, only for `requires_dev_catalog` / artifact-ref deploys.
- **Workflow log streaming for artifact-ref deploys** — per-model refs write logs under `.artifacts/refs/<ref>/workflow_logs/`, which the backend neither mounted nor searched, so the UI log stream 404'd. Adds the read-only mount, includes ref dirs in log discovery, and lazily backfills `workflow_log_path` in `WorkflowLogStreamView` the same way `DeploymentHistoryView` already does.
- **Tool-call parser** — Qwen3.5/3.6 map to the `qwen3_coder` parser their Blackhole builds register; older Qwen families keep `hermes`.
- **Thinking shown inline in the chat UI** — Qwen3.5-9B had no `REASONING_MODELS` entry, so deploys ran without `--reasoning-parser` and its thinking streamed into the visible reply. Maps it to the `qwen3` parser its build registers (mirroring the pinned tt-inference-server spec), which also exposes the `Qwen3.5-9B-thinking` gateway variant. The chat UI additionally restores the opening `<think>` tag when a model emits only the closing one (Qwen-style templates), so leaked thinking still lands in the thinking block on parser-less servers.
- **Auto-placement** — `autoPlacement` claimed the whole board whenever all slots were free, even for models whose placement only allows a card pair; it now respects `placement.allowsFullBoard`.

Dropped from the original PR as already tackled in dev: the 0.19.0 catalog regeneration and Qwen3.5-9B catalog entry (#1210 added a hand-owned entry pinned to a published image), and the dev-mode `.git`-marker fix (landed via #1222). The unused P300x2 card-pair helpers change was dropped as dead code — nothing imports them and dev pins Qwen3.5-9B to P150.

## Testing

- Reasoning-parser mapping and thinking-leak normalization verified: deploy overrides for Qwen3.5-9B now assemble `--tool-call-parser qwen3_coder --reasoning-parser qwen3`; the chat UI's `processContent` routes untagged `…</think>` thinking into the thinking block with tagged/plain/partial-stream cases unchanged. Verified the running gemma-4 deploy streams cleanly through `/models/inference/` with no reasoning leak.
- Parser matrix verified in the running backend container: Qwen3.5/3.6 → `qwen3_coder`; Qwen3-32B/QwQ → `hermes`; gemma-4 → `gemma4`; medgemma → `None`; llama/mistral/deepseek unchanged.
- Refs log discovery verified functionally (exact and fuzzy timestamp match) against a synthetic `.artifacts/refs/<ref>/workflow_logs/docker_server` tree.
- Backend hot-reloaded the changes cleanly; `GET /up/` returns 200. `--disable-metal-timeout` confirmed supported by both the 0.19.0 artifact's `run.py` and the stisi ref's `run.py`.
- `eslint` clean on `deviceFit.ts`; `tsc` failures are pre-existing on dev (`recharts` in `TrainingJobDetailPage.tsx`).

## Remaining validation

- [ ] End-to-end Qwen3.5-9B deploy on hardware with the log stream open (board currently occupied by a gemma-4 deploy)

### DIFF
--- a/app/backend/docker_control/tt_inference_client.py
+++ b/app/backend/docker_control/tt_inference_client.py
@@ -35,6 +35,10 @@ def tool_call_parser_for(model_name: str = "", hf_model_id: str = "") -> Optiona
     s = f"{hf_model_id} {model_name}".lower()
     if "llama-3" in s or "llama3" in s:
         return "llama3_json"
+    # Qwen3.5 / Qwen3.6 blackhole builds ship with the qwen3_coder parser; older
+    # Qwen families use hermes.
+    if "qwen3.5" in s or "qwen3.6" in s or "qwen35" in s or "qwen36" in s:
+        return "qwen3_coder"
     if "qwen" in s or "qwq" in s:
         return "hermes"
     if "mistral" in s:
@@ -136,6 +140,7 @@ def start_chat_deployment(
     override_tt_config: Optional[str] = None,
     override_docker_image: Optional[str] = None,
     artifact_ref: Optional[str] = None,
+    disable_metal_timeout: bool = False,
 ) -> TTInferenceRunResult:
     """Start a chat model deployment via TT Inference Server (/run).
 
@@ -165,6 +170,8 @@ def start_chat_deployment(
         payload["override_docker_image"] = override_docker_image
     if artifact_ref is not None:
         payload["artifact_ref"] = artifact_ref
+    if disable_metal_timeout:
+        payload["disable_metal_timeout"] = True
 
     # Pass UI-managed secrets explicitly. The inference server runs on the host
     # and cannot read user_config.env in the persistent volume when the backend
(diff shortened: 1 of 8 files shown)


jashansinghTT:
all of these issues were addressed in @stisiTT 's base hence closing this for now.

stisiTT (approved):

PR #1220, opened 2026-08-11
feat: add --no-clear flag to keep terminal contents on startup

anirudTT:


jashansinghTT (approved):
Image

From what I understood, it doesn't give all of those initial animations that are present in tt-studio setup in the cli and gives verbose output in a simple format, so it works according to that, I think.

anirudTT:
> Image
> From what I understood, it doesn't give all of those initial animations that are present in tt-studio setup in the cli and gives verbose output in a simple format, so it works according to that, I think.

yes correct thanks for asking that , i dont want it to give those animations

PR #1223, opened 2026-08-12
feat: explain the launcher's image-source decisions and service failures

anirudTT:
Builds on #1188. A failed image pull printed raw daemon errors and then claimed the registry was unreachable when it had answered fine, and a service that didn't come up dumped its log tail under the live step spinner. Both now say what happened and what to do next.

Before:

```
  Image ghcr.io/tenstorrent/tt-studio/backend:sha-205aedf73de2 failed to resolve reference
  "ghcr.io/.../backend:sha-205aedf73de2": not found
  Error response from daemon: failed to resolve reference "…": not found
Registry unreachable — using locally cached images
```

After:

```
  ✓ chroma pulled
  ○ Prebuilt images for sha-205aedf73de2 aren't published — using local images
⠹ Pulling prebuilt images  ▕█████████░░░░░▏  2/3 images · 7.3 MB
```

- [x] Stream `docker compose pull` into one progress line (images completed, bytes downloaded) plus a ✓ per image, instead of raw compose output. Piped compose reports bytes per layer but no layer totals, so the bar counts images — exact — rather than inventing a percentage
- [x] Classify pull failures (unpublished / auth / unreachable / unknown) and state the fallback in one line. An expected failure reads as a note, not an error, and always ends with what happens instead
- [x] Name the frontend setting when a custom config is what forced the local build, so `VITE_APP_TITLE differs from the published build` doesn't have to be guessed at. `frontend_config_is_stock` now also accepts a quoted `"TT Studio"`, which it previously rejected
- [x] Title phase 4 `Pull` or `Build` for what it actually does, retitling if a pull falls back. The phase count stays 5/5
- [x] Diagnose host-service startup failures into a card — cause, one line of evidence, the consequence, then what to try — instead of printing the log tail. Applies to the Docker Control service and the inference server
- [x] Wait for a freed port to actually be released before rebinding. Killing the holder doesn't hand the port back instantly and `uvicorn --reload` passes its socket to a child, which is what made the Docker Control service fail to restart with `[Errno 98] Address already in use`
- [x] Keep background services' own output off the terminal (`DEVNULL`, they already log to their files) — a subprocess inherits fd 1, so `redirect_stdout` never reached it and its errors landed on the spinner's row
- [x] Fold two double-confirmations and dedupe compose's repeated `Container … Healthy` lines
- [x] Add a `cli-design` skill: the launcher's design language packaged for other repos, with a runnable CLI skeleton, the console module, a demo, a test file to copy, and a one-command installer

Rebased onto the base branch after f0f6bfd landed there, which fixes the same `Built` vs `Started` mislabeling this branch had also addressed — upstream's version is kept and the duplicate dropped.

Verification:

- [x] 268 tests pass (`.tt_studio_run_venv/bin/python -m unittest discover -s tests -q`), 45 of them new — parsers, classifiers, fallback wording, port release, phase titles
- [x] Real `python run.py --dev` run: ready in 13.6s, Docker Control starts cleanly where it previously failed
- [x] Render-checked under a PTY at 80/90/100/110 columns, and piped through `cat -v` to confirm non-TTY output carries no escape codes
- [x] Failure paths reproduced against a real registry (nonexistent tag, occupied port) rather than mocked

Worth a separate discussion: `publish-images.yml` only fires on `v*` tags, releases, and manual dispatch, so a `sha-<12>` tag is never published and any non-tag checkout takes the local-build path. The prebuilt images currently help only checkouts sitting exactly on a release tag.


jashansinghTT (approved):
Works as intented

PR #1228, opened 2026-08-13
test: add mochi-1-preview on the patched 0.18.0 media image (QB2)

jashansinghTT:
Testing branch for mochi-1-preview on Blackhole QuietBox 2 while the upstream fixes land.

Background: mochi needs three fixes, all upstream in tt-inference-server — a newer media image (the artifact pins 0.10.0-555f240, whose tt-metal rejects the P300x2 2x2 mesh), `TT_DIT_CACHE_DIR` in the container env, and the runner-side pipeline call fix (tenstorrent/tt-inference-server#4940, spec pin in #4941). Until a release carries those, this branch runs mochi on a patched image: `ghcr.io/tenstorrent/tt-studio/studio_images:mochi-1-preview-qb2-20260813-0.18.0-c49bb76` (stock 0.18.0 + the #4940 diff layered on top).

Rebased onto dev and reworked to use the declarative overrides in `model_overrides.toml` instead of code and a hand-edited catalog:

- [x] Drop the `[[unavailable]]` mark for mochi on P300x2
- [x] Add a `[[serve_override]]` pinning p300x2 deploys to the patched image. Scoped to p300x2, the only board it was verified on; other boards keep the spec image as before
- [x] Pull and deploy agree: dev's `media_image_override()` feeds both `run_container` and the media pre-pull, so the stale 0.10.0 spec image is no longer pulled
- [x] `TT_DIT_CACHE_DIR`: `[[serve_override]]` now takes `env_vars`. The sync script derives a runtime model spec for that model/device from the real upstream spec with the env merged in (and the image pin baked in, since run.py loads a runtime spec JSON as-is) and points `runtime_model_spec_overrides` at it — the same channel the chip-tier overrides use, so `run_container` needs no new code. mochi gets `/home/container_app_user/cache_root/tt_dit_cache`, the path the Wan2.2 spec uses
- [x] Catalog regenerated with `sync_models_from_inference_server.py` from the 0.21.0 artifact (no hand edits); the only changes are P300x2 rejoining mochi's `device_configurations` and its `runtime_model_spec_overrides` entry. `--resync` is safe to use
- [x] Unit tests: the pin (`MediaImageOverrideTests`), the `/run` payload `run_container` posts for mochi vs an unpinned media model (`RunContainerMediaPinTests`), and the spec derivation (`TestServeEnvOverrides`). `test_sync_models.py` / `test_model_config.py` otherwise match the clean-dev baseline. The derived spec also loads through the artifact's `ModelSpec.from_json`
- [x] The patched image itself was verified on a QB2 on the original branch: deploy → warmup → 12-step generation → valid MP4
- [ ] Hardware re-verification of the env override: not yet run. The box was in use for another PR when this landed, so the `TT_DIT_CACHE_DIR` fix is covered by unit tests and the spec load only

Steps to test:
1. `git fetch && git checkout jashan/mochi-only-spec-test`
2. The patched image must be present: `docker pull ghcr.io/tenstorrent/tt-studio/studio_images:mochi-1-preview-qb2-20260813-0.18.0-c49bb76` (the package is public — no login needed)
3. `python run.py --dev` (the sync writes `app/backend/shared_config/runtime_model_specs/mochi-1-preview-p300x2.json`; check it carries `TT_DIT_CACHE_DIR` and the studio_images tag)
4. Open http://localhost:3000 → deploy — mochi-1-preview should appear alongside the existing models; deploy it on the full board (all 4 chips). The inference server log should show `--runtime-model-spec-json` and the container env should include `TT_DIT_CACHE_DIR`
5. Warmup takes ~3 min after image load; the container is healthy when `GET :8000/models/health/?deploy_id=<id>` returns Healthy
6. Generate a video from the playground (or `POST /models/video-generation/` with `{"deploy_id": ..., "prompt": ..., "num_inference_steps": 12}`), then download — expect a valid MP4

Drop the image pin and env override once an inference-server artifact release carries #4940/#4941; the real integration is then just the artifact version bump.


rnabeelTT (commented):
This also suffers from the same issue: pulls inference-media-serverv0.10.0 image unnecessarily. Deployment also fails with:
2026-08-18 17:47:29,660 - ERROR - Device 0,1,2,3: Mochi pipeline creation failed
Traceback (most recent call last):
  File "/home/container_app_user/tt-metal/server/tt_model_runners/dit_runners.py", line 301, in create_pipeline
    return MochiPipeline.create_pipeline(
  File "/home/container_app_user/tt-metal/models/tt_dit/pipelines/mochi/pipeline_mochi.py", line 198, in create_pipeline
    return cls(device=mesh_device, config=config)
  File "/home/container_app_user/tt-metal/models/tt_dit/pipelines/mochi/pipeline_mochi.py", line 234, in __init__
    raise RuntimeError(msg)
RuntimeError: Cache must be enabled when DiT model reloading is enabled (reload_dit_model=True). Please set TT_DIT_CACHE_DIR environment variable to enable caching.

jashansinghTT:
Both review points are addressed as of 5a7a41c, on top of the rebase onto dev.

**Unnecessary pull of the 0.10.0 image.** The pin moved into `model_overrides.toml` as a `[[serve_override]]`, and dev's `media_image_override()` is consulted by both the media pre-pull in `views.py` and `run_container`, so the image the UI shows progress for is the studio_images build that actually deploys. `MediaImageOverrideTests` covers the pin.

**`TT_DIT_CACHE_DIR` / "Cache must be enabled when DiT model reloading is enabled".** The patched runner needs the cache dir and the mochi spec carries no env_vars. There was no studio-side way to add one (catalog `env_vars` are not sent to media deploys, run.py has no env override flag), so `[[serve_override]]` now takes `env_vars`. The sync script derives a runtime model spec for that model/device from the real upstream spec with the env merged in — the same channel the chip-tier overrides use — and `run_container` deploys it via `--runtime-model-spec-json`. Since run.py loads a runtime spec as-is and skips `apply_overrides`, the derived file also carries the image pin. mochi gets `/home/container_app_user/cache_root/tt_dit_cache`, the path the Wan2.2 spec uses, so the cache persists across redeploys.

On the Copilot notes: the catalog is generated, so the env var lives in the TOML entry and lands in the derived spec rather than in `models_from_inference_server.json`; and `RunContainerMediaPinTests` now asserts the `/run` payload `run_container` posts for mochi (pinned image + derived spec, no `device_id`) against an unpinned media model.

Unit-tested (`TestServeEnvOverrides`, `RunContainerMediaPinTests`, the derived spec loaded through the artifact's `ModelSpec.from_json`). Not yet re-deployed on hardware — the QB2 was in use for another PR — so a redeploy on P300x2 is the remaining check.

rnabeelTT:
Model stalls for me during deployment

PR #1230, opened 2026-08-14
fix: give the agent a model token on fresh installs

### LINKED ISSUE #1229: Agent service fails to start on fresh install or after purge: no JWT available
## What happens

On a fresh install, or after `--purge-all`, the agent service never starts up. It logs this on its first line and then retries forever:

```
Warning: No authentication token available (neither CLOUD_CHAT_UI_AUTH_TOKEN nor JWT_SECRET)
Auto-discovery failed: 1 validation error for CustomLLM
ERROR: No LLM endpoint available
```

Model discovery itself works. The agent finds a healthy chat model and builds the right URL. It only fails on the last step, constructing the client, because it has no token. `CustomLLM.encoded_jwt` is a required string and gets `None`.

## Why

`JWT_SECRET` is commented out in `.env.default` since #750 moved it to UI-managed storage. The backend copes because `get_jwt_secret()` falls back to the stored config in the persistent volume and auto-generates one. The agent has no such fallback. It only reads the environment variable, which is empty.

Existing installs still work because their `.env` predates #750 and carries the value. Any purge or fresh clone hits this.

## Why it looks already fixed

#750 did implement the wiring, in `run_agent_container()` in `docker_control/docker_utils.py`. That function passes `JWT_SECRET`, passes `INTERNAL_PERSISTENT_STORAGE_VOLUME`, and mounts the persistent volume read-only.

`run_agent_container()` has no callers. Repo-wide search finds only the definition. The last commits touching it are from #231 (2025-07-01) and #168 (2025-02-07). The agent that actually runs is the compose service `tt_studio_agent`, which has none of that wiring.

So the mechanism reads as present to anyone grepping for it, which is likely why this went unnoticed.

## Side effect

The same dead function is why the Tavily hot-swap from #750 has never worked. `_resolve_tavily_api_key()` starts with `base = os.environ.get("INTERNAL_PERSISTENT_STORAGE_VOLUME")`, which is never set on the compose agent, so the UI-store lookup is skipped and it silently falls back to the environment. Changing the key in the Settings UI does not affect a running agent.

## Notes on fixing

Mounting the secrets file into the agent is not sufficient on its own. `user_config.env` is written `0600` and root-owned, and the compose agent runs as `1000:1000`, so it cannot read it. The existing Tavily reader is already guarded for this and gives up silently.

Putting `JWT_SECRET` back in `.env` is also risky. `get_jwt_secret()` prefers the stored config over the environment, so on an install that already has a generated secret the agent and backend would end up using different values.

The backend is the only component that reliably knows the resolved secret, so having it hand the agent a token avoids duplicating the secret, the file permissions problem, and any precedence change.

Worth deleting `run_agent_container()` as part of this, since leaving it is what made the problem look solved.

stisiTT:
Closes #1229

## What was broken

On a fresh install, or after `--purge-all`, the agent never started. It printed "No authentication token available" on its first line and then retried forever.

It needs a token to talk to a deployed model, and it had no way to get one. `JWT_SECRET` has been optional in `.env` since #750, and when it is unset the backend generates one into the UI-managed config instead. The backend reads it from there. The agent only ever read the environment variable, so on a clean setup it got nothing.

Existing machines kept working because their `.env` predates #750 and still carries the value, which is why this went unnoticed.

## The fix

The backend is the only component that reliably knows the secret, so it now signs a token on request and the agent asks for one. A new `GET /models/agent/auth-token/` reuses the existing token helper.

The agent resolves it lazily, on first use, rather than at import. That detail matters: `poll_for_llm()` retries but reuses the module global, and `setup_discovered_llm()` returns the LLM even when the connection test fails, so a one-shot fetch at import would turn a brief backend outage into an agent that reports itself initialized and then 401s every message. Caching only on success lets the existing retry loop recover on its own.

Two alternatives were rejected:

- **Mounting the stored config into the agent.** The file is written `0600` and root-owned, and the agent runs as a non-root user, so it cannot read it.
- **Putting `JWT_SECRET` back in `.env`.** The backend prefers the stored value, so the two containers would end up using different secrets.

Cloud token and an explicit `JWT_SECRET` keep priority, in that order. The env branch is checked before the backend mint on purpose: `inference-api` resolves `JWT_SECRET` env-first when launching a model, so `.env` winning here matches the secret the model was actually started with.

## Verified

The mechanism was confirmed against the running stack before it was written. A token signed with the backend's resolved secret, which is exactly what the new endpoint returns, authenticated to a deployed Qwen3.5-9B and returned a completion.

The retry behaviour was checked separately against a simulation of the caching contract: a failed fetch is not cached, a later attempt succeeds once the backend is back without needing a restart, and no further requests are made after success.

Also checked: all three files compile, no construction site still reads the module global, there is no longer any network call at import time, and the URL prefix matches the agent's existing `/models/deployed/` call.

## Not covered

A registered external container running its own secret still needs the per-model token the backend builds separately. Only the shared-secret case goes through this endpoint. Noted in #1229.

The same missing wiring is also why the Tavily hot-swap from #750 has never worked on a running agent. Separate fix, also noted in #1229.

Removing the dead `run_agent_container()`, which is what made this bug look already solved, is split into #1231 to keep this PR to the fix.

## Test plan

- [ ] With `JWT_SECRET` unset in `.env`, start fresh and confirm the agent reports requesting a token from the backend rather than the warning
- [ ] Send a message through the agent and confirm it reaches the model
- [ ] Stop the backend, restart the agent, confirm it keeps retrying and then recovers once the backend is back, rather than settling into a healthy-but-broken state
- [ ] With `JWT_SECRET` set in `.env`, confirm that path still takes precedence
- [ ] In cloud mode, confirm the cloud token still wins

### DIFF
--- a/app/agent/agent.py
+++ b/app/agent/agent.py
@@ -119,7 +119,37 @@ json_payload = json.loads('{"team_id": "tenstorrent", "token_id":"debug-test"}')
 jwt_secret = os.getenv("JWT_SECRET")
 cloud_auth_token = os.getenv("CLOUD_CHAT_UI_AUTH_TOKEN")
 
-# Use cloud auth token if available, otherwise fall back to JWT
+
+def _fetch_backend_auth_token() -> str:
+    """Ask the backend to mint a token for calling a deployed model.
+
+    JWT_SECRET is optional in .env, and when unset the backend keeps the secret in
+    the UI-managed config, which this container cannot read. Returns "" on failure
+    so a missing token surfaces as a 401 rather than a validation error.
+    """
+    try:
+        resp = requests.get(f"{AgentConfig.BACKEND_URL}/models/agent/auth-token/", timeout=10)
+        if resp.status_code == 200:
+            return resp.json().get("token") or ""
+        print(f"Backend returned {resp.status_code} for agent auth token")
+    except Exception as e:
+        print(f"Could not fetch agent auth token from backend: {e}")
+    return ""
+
+
+def _auth_token() -> str:
+    """Token for local model calls, minted by the backend on first use and cached.
+
+    Resolved lazily so poll_for_llm()'s retries can recover if the backend was not
+    reachable yet; fetching at import would leave the agent permanently tokenless.
+    """
+    global encoded_jwt
+    if not encoded_jwt:
+        encoded_jwt = _fetch_backend_auth_token()
+    return encoded_jwt
+
+
+# Cloud token wins, then an explicit JWT_SECRET, else the backend mints one on demand.
 if cloud_auth_token:
     encoded_jwt = cloud_auth_token
     print(f"Using cloud auth token for authentication")
@@ -127,8 +157,8 @@ elif jwt_secret:
     encoded_jwt = jwt.encode(json_payload, jwt_secret, algorithm="HS256")
     print(f"Using JWT authentication")
 else:
-    encoded_jwt = None
-    print("Warning: No authentication token available (neither CLOUD_CHAT_UI_AUTH_TOKEN nor JWT_SECRET)")
+    encoded_jwt = ""
+    print("JWT_SECRET not set; will request a model token from the backend")
 
 class RequestPayload(BaseModel):
     message: str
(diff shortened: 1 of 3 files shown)


jashansinghTT (approved):
Made the JWT reachable and is a good fix

PR #231, opened 2025-03-10
Anirud/ai playground

anirudTT:

PR #1231, opened 2026-08-14
chore: remove the unreachable run_agent_container

stisiTT:
Split out of #1230 so that PR stays a focused bug fix. Context in #1229.

## Why

`run_agent_container()` is why the missing agent-token wiring looked like it was already solved. #750 added the `JWT_SECRET` pass-through, `INTERNAL_PERSISTENT_STORAGE_VOLUME` and the persistent-volume mount inside this function, so the mechanism reads as present to anyone grepping for it.

Nothing calls it. A repo-wide search finds only the definition, and the last commits touching it are #231 and #168, both from 2025.

It could not have worked anyway: it called `get_host_agent_port()`, which is not defined anywhere in the tree, so the first call would have raised `NameError`. That makes it unreachable rather than merely unused.

The agent that actually runs is the compose service, which #1230 fixes by having it request a token from the backend.

## Also removed

Deleting the function leaves `get_tavily_api_key` imported but unused. ruff F401 runs via `dev-tools/.pre-commit-config.yaml`, so leaving it would surface on someone else's unrelated commit.

`copy` and `backend_config` both have other callers and are untouched.

## Checked

- No remaining references to `run_agent_container` or `get_host_agent_port` anywhere under `app/`
- `get_tavily_api_key` still has 5 callers elsewhere in the backend
- File compiles

Pure deletion, no behaviour change.

### DIFF
--- a/app/backend/docker_control/docker_utils.py
+++ b/app/backend/docker_control/docker_utils.py
@@ -15,7 +15,6 @@ from shared_config.logger_config import get_logger
 from shared_config.model_config import model_implmentations, _impl_selector
 from shared_config.backend_config import backend_config
 from shared_config.model_type_config import ModelTypes
-from shared_config.user_config import get_tavily_api_key
 from shared_config.coding_agent_config import is_coding_agent_eligible
 from board_control.services import SystemResourceService
 from docker_control.models import ModelDeployment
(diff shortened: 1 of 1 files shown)


jashansinghTT (approved):
Removal here is valid because this function was never used

PR #1233, opened 2026-08-14
feat: add qwen3.8-27b for p300x2

stisiTT:
Makes Qwen3.8-27B deployable and usable in the chat UI on a P300x2 board (4 chips).

The model isn't in the tt-inference-server build TT-Studio normally ships with, so the catalog entry points at its own build and its own image instead of the default one. The image is published to `tt-studio/studio_images`, so this works on any box and not just the one it was built on.

Two lookups in the backend also needed the new version. The first picks the vLLM tool-call parser by matching on the model name, and it didn't know about "3.8". That meant the model fell through to `hermes`, and TT-Studio then sent that as an override which replaced the `qwen3_coder` parser the model actually asks for. The second controls which models get a thinking mode. Without an entry there, the model got no thinking variant and no reasoning stats even though it does produce reasoning.

- [x] Add the catalog entry, P300x2 only so it allocates all 4 chips
- [x] Recognize 3.8 in the tool-call parser lookup so it resolves to `qwen3_coder` instead of `hermes`
- [x] Register it as a reasoning model so it gets a thinking variant
- [x] Deployed on a P300x2 box, pulled the image from GHCR, confirmed healthy
- [x] Checked the container starts with `--tool_call_parser qwen3_coder` and `--reasoning_parser qwen3`
- [x] Chatted through the same inference path the UI uses, and the thinking came back separately from the answer rather than leaking into the reply
- [x] Confirmed the weights load from the normal Hugging Face cache with no re-download

Marked EXPERIMENTAL. Vision isn't reachable from the UI yet, so it's text only in practice, though the server still warms the vision tower on start.

### DIFF
--- a/app/backend/docker_control/tt_inference_client.py
+++ b/app/backend/docker_control/tt_inference_client.py
@@ -35,9 +35,12 @@ def tool_call_parser_for(model_name: str = "", hf_model_id: str = "") -> Optiona
     s = f"{hf_model_id} {model_name}".lower()
     if "llama-3" in s or "llama3" in s:
         return "llama3_json"
-    # Qwen3.5 / Qwen3.6 blackhole builds ship with the qwen3_coder parser; older
-    # Qwen families use hermes.
-    if "qwen3.5" in s or "qwen3.6" in s or "qwen35" in s or "qwen36" in s:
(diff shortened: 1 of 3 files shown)


jashansinghTT (approved):
Image

This works well for all the operations at the present.

rnabeelTT (approved):
Works well! tested through chat UI, and tool calling tested through OpenCode!

PR #1234, opened 2026-08-14
feat: add motif-image-6B-preview for p300x2

jashansinghTT:
The v0.21.0 artifact lists Motif-Image-6B-Preview on P300X2, but no published media image can actually run it there: the spec's pinned 0.9.0-c180ef7 rejects the model name outright (stale lowercase enum, no P300X2 device config), and the newest stock 0.18.0 image still lacks a (2, 2) mesh preset in tt-metal plus enough trace region for the tp=2 layout. Dev currently hides the pair with an `[[unavailable]]` mark.

Rebased onto dev and reworked to use the declarative overrides in `model_overrides.toml` instead of code and a hand-edited catalog:

- [x] Drop the `[[unavailable]]` mark for Motif on P300x2
- [x] Add a `[[serve_override]]` pinning p300x2 deploys to the patched image `ghcr.io/tenstorrent/tt-studio/studio_images:motif-image-6b-p300x2-20260814-0.18.0-c49bb76` (stock 0.18.0-c49bb76 plus the two one-line fixes, tt-metal#53197 and tt-inference-server#4955). Scoped to p300x2, the board the patch targets and the only one verified; other boards keep the spec image as before
- [x] Regenerate the catalog with `sync_models_from_inference_server.py` from the 0.21.0 artifact (no hand edits) — the only catalog change is P300x2 rejoining Motif's `device_configurations`
- [x] Pull and deploy agree: dev's `media_image_override()` already feeds both `run_container` and the media pre-pull, so the earlier review point about pulling the spec image needlessly is covered by dev's design
- [x] Unit tests for the pin (`MediaImageOverrideTests`), plus `test_sync_models.py` / `test_model_config.py` match the clean-dev baseline
- [x] Hardware verification on p300x2 was done on the original branch: UI deploy flow, warmup healthy, two 1024x1024 generations in ~7s each. The rebased change is data-only, so it was not redeployed on hardware
- [x] Patched image is already in GHCR

Drop the override once a media image ships with both fixes and the spec re-pins (tt-inference-server#4956).

### DIFF
--- a/app/backend/docker_control/docker_utils.py
+++ b/app/backend/docker_control/docker_utils.py
@@ -423,6 +423,18 @@ def run_container(impl, weights_id, device_id=0, host_port=None, use_image_overr
         if impl.model_name in {"Wan2.2-T2V-A14B-Diffusers"}:
             payload["override_docker_image"] = "ghcr.io/tenstorrent/tt-media-inference-server:0.17.0-8c48a10"
 
+        # Motif pinned to a patched 0.18.0 media image. The artifact spec pins
+        # 0.9.0-c180ef7, which cannot run Motif at all: its ModelNames enum predates
+        # the rename to exact-case values ("motif-image-6b-preview" vs the
+        # "Motif-Image-6B-Preview" that run.py passes as MODEL) and it has no Motif
+        # P300X2 device config. The stock 0.18.0 image fixes both but its tt-metal
+        # MotifPipeline lacks a (2, 2) mesh preset, so the p300x2 mesh dies with
+        # KeyError: 'num_links'. The patch layer adds that preset (T3K preset with
+        # tensor-parallel halved); drop it once tt-metal ships the preset and the
+        # artifact re-pins.
+        if impl.model_name == "Motif-Image-6B-Preview":
+            payload["override_docker_image"] = "ghcr.io/tenstorrent/tt-studio/tt-media-inference-server:0.18.0-c49bb76-motif-p300x2"
+
         # Disambiguate the target model_spec. Some models share a name+device across
         # engines (e.g. Llama-3.1-8B has both a vLLM chat spec and a forge training
         # spec on P150); without an impl the server defaults to the wrong engine and
--- a/app/backend/docker_control/docker_utils.py
+++ b/app/backend/docker_control/docker_utils.py
@@ -433,7 +433,7 @@ def run_container(impl, weights_id, device_id=0, host_port=None, use_image_overr
         # tensor-parallel halved); drop it once tt-metal ships the preset and the
         # artifact re-pins.
         if impl.model_name == "Motif-Image-6B-Preview":
-            payload["override_docker_image"] = "ghcr.io/tenstorrent/tt-studio/tt-media-inference-server:0.18.0-c49bb76-motif-p300x2"
+            payload["override_docker_image"] = "ghcr.io/tenstorrent/tt-studio/studio_images:motif-image-6b-p300x2-20260814-0.18.0-c49bb76"
 
         # Disambiguate the target model_spec. Some models share a name+device across
         # engines (e.g. Llama-3.1-8B has both a vLLM chat spec and a forge training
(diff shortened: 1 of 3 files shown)


rnabeelTT (commented):
### Problem: we are wastefully pulling both the patched and the original image

What's happening
Motif's pin lives in one place only — a hardcoded block inside run_container:

[docker_utils.py:435-436](vscode-webview://1bktm79uf2r8k97avltdqd58t3jvgk9jv3pu3kuiddr092a54crm/app/backend/docker_control/docker_utils.py#L435-L436) sets payload["override_docker_image"] = ".../studio_images:motif-image-6b-p300x2-...-0.18.0-c49bb76".

But the pre-pull that runs before that, on the media/forge branch, never consults it:

[views.py:889](vscode-webview://1bktm79uf2r8k97avltdqd58t3jvgk9jv3pu3kuiddr092a54crm/app/backend/docker_control/views.py#L889)


deploy_image = resolve_deploy_image(impl.model_name, media_device, impl=inference_impl) or impl.image_version
resolve_deploy_image asks inference-api GET /resolve-image, which answers from the tt-inference-server model_spec — and that spec is exactly what pins 0.9.0-c180ef7 (your own comment at [docker_utils.py:427](vscode-webview://1bktm79uf2r8k97avltdqd58t3jvgk9jv3pu3kuiddr092a54crm/app/backend/docker_control/docker_utils.py#L427) says so). Since it returns a value, the or impl.image_version fallback to the catalog's correct docker_image never fires. So the UI pulls 0.9.0-c180ef7 with a nice progress bar, then run_container deploys the studio image — which is likely not cached, so the real pull happens silently inside the inference server afterward.

Net effect: a wasted multi-GB pull of an image that can't run Motif, and pull progress that doesn't correspond to the image being deployed. The catalog JSON is not being ignored at deploy time — it's just not consulted at pull time either; the override is what saves the deploy.

Why Qwen3.5-9B is fine
It's a CHAT model, so it goes through the vLLM branch, where the same value feeds both steps:

[views.py:709](vscode-webview://1bktm79uf2r8k97avltdqd58t3jvgk9jv3pu3kuiddr092a54crm/app/backend/docker_control/views.py#L709) → override_docker_image = _resolve_override_docker_image(impl), which for requires_dev_catalog: true returns impl.image_version — the catalog's studio_images:qwen35-9b-... pin — and then [views.py:732](vscode-webview://1bktm79uf2r8k97avltdqd58t3jvgk9jv3pu3kuiddr092a54crm/app/backend/docker_control/views.py#L732):


deploy_image = override_docker_image or resolve_deploy_image(...) or impl.image_version
override_docker_image is checked first, so pre-pull and deploy agree. Two differences compound for Motif: its override is hardcoded in docker_utils instead of derived from the catalog by _resolve_override_docker_image, and the media pre-pull line lacks the override_docker_image or prefix. Wan2.2-T2V has the identical latent bug at [docker_utils.py:424](vscode-webview://1bktm79uf2r8k97avltdqd58t3jvgk9jv3pu3kuiddr092a54crm/app/backend/docker_control/docker_utils.py#L424) — you just may not have noticed if its spec image happened to be cached.

Fix
Lift the per-model media pins out of run_container into a module-level resolver (e.g. _MEDIA_IMAGE_OVERRIDES = {"Motif-Image-6B-Preview": ..., "Wan2.2-T2V-A14B-Diffusers": ...} with a resolve_media_override_image(impl) helper), have run_container read it, and prefix the media pre-pull the same way the chat path does:


deploy_image = (resolve_media_override_image(impl)
                or resolve_deploy_image(impl.model_name, media_device, impl=inference_impl)
                or impl.image_version)

jashansinghTT:
Addressed the review in 2625bae: lifted the Wan2.2/Motif pins out of `run_container` into a module-level `_MEDIA_IMAGE_OVERRIDES` map with a `resolve_media_override_image(impl)` helper, and the media pre-pull in `views.py` now checks it first — `resolve_media_override_image(impl) or resolve_deploy_image(...) or impl.image_version` — mirroring the chat path's override-first ordering. Pull and deploy now agree for both Motif and Wan2.2, so the stale spec-pinned image is no longer pulled.

PR #1236, opened 2026-08-16
fix: suppress pip root-user warnings in Docker builds

### LINKED ISSUE #959: Suppress pip root-user warnings in Docker builds
## What's happening

When building our Docker images, pip prints this warning a bunch of times during the build:

> WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the `--root-user-action` option if you know what you are doing and want to suppress this warning.

It's noisy, clutters the build output, and makes it harder to spot real issues. Inside our containers we intentionally install as root (no venv needed in an isolated container), so the warning is not actionable.

## Fix

Add `--root-user-action=ignore` to the pip commands in our Dockerfiles. Quick, low-risk change.

## Files to update

- `app/backend/Dockerfile` — has multiple `pip3 install` calls running as root that need the flag:
  - The tt-smi install block (`pip3 install --upgrade pip` and `pip3 install .`)
  - The requirements install block (`pip3 install --upgrade pip` and `pip3 install -r requirements.txt`)
  - The openwakeword install (`pip3 install --no-cache-dir --no-deps openwakeword==0.6.0`)
- `app/agent/Dockerfile` — single `pip install -r requirements.txt --no-cache-dir` line that runs as root.
- `models/dummy_echo_model/Dockerfile` — this one already switches to a non-root `user` before running `pip install`, so it shouldn't be emitting the warning. Worth a quick check during the PR but likely no change needed.

## Acceptance

- `docker build` for backend and agent images no longer shows the pip root-user warning.
- No change in installed packages or build behavior otherwise.

AnkittChauhan:
### Summary
Suppresses the noisy `WARNING: Running pip as the 'root' user...` warnings when building backend and agent Docker images.
- Fixes #959

### What changed
- Added `ENV PIP_ROOT_USER_ACTION=ignore` to `app/backend/Dockerfile` and `app/agent/Dockerfile`.
- Using the environment variable covers all current and future `pip install` commands in both Dockerfiles without needing repetitive `--root-user-action=ignore` flags on every `RUN` line.
- Checked `models/dummy_echo_model/Dockerfile` — it already runs as a non-root user inside a venv, so no changes were needed there.

### Testing
- Built both `backend` and `agent` images locally; verified the root-user warnings are gone from build logs.
- Verified dependencies still install normally with no change in build or runtime behavior.


jashansinghTT (approved):
Thanks for suppressing the warnings. This LGTM @AnkittChauhan . Also thanks for your contributions like always.

PR #1238, opened 2026-08-17
feat: add custom dataset upload, preview, and management for training

mcupacTT:
Lets users upload JSON datasets from the Training page, preview them in a table/JSON view, and list or delete previously uploaded datasets. Uploaded datasets are stored on the shared training volume and offered as choices in the New Training Job dialog.


mcupacTT:
closing since this was a tryout to separate a big PR into two smaller - but we managed to merge the big one so this one is not necessary

PR #1245, opened 2026-08-18
fix: focus onboarding on the hugging face token

jashansinghTT:
Follow-up to the onboarding feedback in Slack. The setup wizard front-loaded too much: auto-managed values the user can't edit, a third-party search key, and an access check that implied public models are gated.

- [x] Secrets step now asks only for the HF token; the auto-managed TTS API key and JWT secret are no longer shown, and the Tavily key is managed from the Settings dialog instead
- [x] HF token copy leads with prioritized weight downloads, with gated-model access as the secondary benefit
- [x] Access check only lists actually gated repos (Llama 3.1, Llama 3.3, FLUX.1-dev); Qwen3-32B and Wan2.2 are public and were removed from both the backend list and the frontend placeholder (verified against the HF API)
- [x] Removed em dashes from wizard copy
- [x] AgentStatusView reads the Tavily key via user_config instead of the container-start env, so a key added later in Settings enables the search agent without a backend restart

Verified with a scripted browser walkthrough of all four wizard steps (single HF field, three access rows, copy sweep), frontend type-check/lint/build, backend api + user_config test suites in a throwaway backend container (same pass/fail set as origin/dev; the failures are pre-existing harness gaps), and a live check that a Tavily key saved via user_config flips tavily_configured with no env var set.

### DIFF
--- a/app/backend/api/hf_access.py
+++ b/app/backend/api/hf_access.py
@@ -11,12 +11,12 @@ import requests
 from typing import List, Dict, Optional
 
 
-# diffusers repos (Wan/FLUX) have no root config.json — check model_index.json instead
+# Only repos that are actually gated on Hugging Face belong here (Qwen and Wan
+# are public). Diffusers repos (FLUX) have no root config.json, so check
+# model_index.json instead.
 HF_GATED_MODELS = [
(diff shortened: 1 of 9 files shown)


rnabeelTT (approved):
LGTM

PR #1246, opened 2026-08-18
fix: fix stale model dropdown after reset, warmup clock reset on refresh, and fake video eta

jashansinghTT:
Three UI-feedback fixes around deploy and monitoring flows.

After a board reset, the model dropdown kept showing "in use by X" until a page refresh — the chip-status poller never listened to the reset dialog's refresh signals. The warmup card's elapsed clock was counted from component mount, so refreshing the page reset it to 0:00. And video generation paced its progress bar against a hardcoded ~4:15 estimate that had nothing to do with the real generation time.

- [x] Refetch chip status when a board reset completes, so the dropdown updates immediately (only the chip-status data refreshes, nothing else)
- [x] Add server-computed `elapsed_seconds` (from the deploy record's `deployed_at`) to the startup-phase payload and anchor the warmup clock to it; falls back to the old mount-based counter if absent
- [x] Drop the hardcoded 255s video estimate: show the true elapsed time from submit with an activity bar, and keep displaying the video the moment the job completes

Verified with the dev stack up (backend/frontend health checks green): the elapsed computation was checked against the live deploy cache in the backend container, and each flow was walked through in a scripted browser session against the real frontend with mocked API responses — dropdown clears after reset without a reload, warmup clock resumes at the true elapsed time across a page refresh, and the video appears as soon as the status flips to completed with no fake countdown. A real board reset / video-model deploy wasn't exercised since the box's hardware was occupied by a live deployment. Frontend lint and type-check pass on the touched files.

### DIFF
--- a/app/backend/model_control/views.py
+++ b/app/backend/model_control/views.py
@@ -359,6 +359,22 @@ class ModelHealthView(APIView):
                 # the container's stdout. Best-effort: if anything fails we still
                 # return the basic 202 so the badge logic is unaffected.
                 content["phase"] = _get_startup_phase(deploy_id)
+                # Anchor the frontend's warmup clock to the deploy start time so
+                # it survives page refreshes. Server-computed to avoid client
+                # clock skew; omitted if deployed_at is missing/unparseable.
+                if content["phase"] is not None:
+                    deployed_at = deploy.get("deployed_at")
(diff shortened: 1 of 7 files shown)


rnabeelTT (approved):
LGTM

PR #1247, opened 2026-08-19
Rc v2.10.0

jashansinghTT:
# Rc v2.10.0

Cut from `main`; validated changes cherry-picked from `dev` (see CONTRIBUTING → Release Process).

## TT Studio v2.10.0 — RC Test Plan

Host: **p300x2 (QB2)**. Check a box only when the criterion passes end-to-end. This plan covers **(1) deployment of every P300x2 model** and **(2) features new since v2.9.0** — earlier functionality is assumed already validated in prior RC plans.

### P300x2 Model Deployment

For each model: deploy from the UI, confirm the capacity-aware deploy flow and image-pull → container-start progress, and confirm the model appears in the navbar only once healthy. Then run the type-specific smoke check.

#### Chat / LLM

- [x] **Llama-3.1-8B-Instruct** deploys, becomes healthy, and responds in chat.
- [x] **Llama-3.3-70B-Instruct** deploys (whole-board) and responds in chat.
- [x] **Qwen3-32B** deploys; the thinking/reasoning toggle works and reasoning tokens render correctly.
- [x] **gemma-4-31B-it** deploys and responds in chat.
- [x] **Qwen3.6-27B** deploys and responds in chat.
- [x] **Qwen3.8-27B** deploys and responds in chat.

#### Image generation

- [x] **FLUX.1-dev** deploys and generates an image from a prompt.
- [x] **FLUX.1-schnell** deploys and generates an image from a prompt.
- [x] **Z-Image-Turbo** deploys and generates an image from a prompt.

#### Video generation

- [x] **Wan2.2-T2V-A14B-Diffusers** deploys and generates a video from a prompt; the ETA shown is reasonable (not a fake/stuck estimate).

#### Speech & audio

- [x] **whisper-large-v3** deploys and transcribes audio; example API snippets load.
- [x] **distil-large-v3** deploys and transcribes audio.
- [x] **speecht5_tts** deploys and produces audio; media-model download progress shows.

### Launcher & images

- [x] Fresh clone + `python run.py`: images are **pulled from GHCR** (not built), and the startup output explains the image-source decision for each service (#1188, #1223)
- [x] `python run.py --build-images` still builds locally and is picked over GHCR
- [x] `python run.py --no-clear` starts without clearing the terminal and streams full setup detail (#1220)
- [x] `python run.py --purge-model` with no value opens the model picker; with a model name it uninstalls just that model, and a shared image is only removed when the last model using it is purged (#1207)
[...]

PR #1248, opened 2026-08-19
ci: allow release PR titles in commitlint check

anirudTT:
The PR title check rejects release PRs. `rc-* -> main` PRs are titled `Rc vX.Y.Z` (what `gh pr create --fill` produces from the branch name), which has no conventional type or subject, so it fails — see #1247. The check landed after the last few releases, so release PRs were never accounted for.

This exempts release branches from the check and adds a `release` type for anyone who'd rather use a conventional title. The two skills and CONTRIBUTING that tell people how to title a PR predate the check too, so they're updated to match what CI actually enforces.

- [x] Skip the title check for `rc-*` and `tt_qb2_launch_branch` head branches (still inside the same job, so the check reports green rather than disappearing)
- [x] Add `release` to commitlint's `type-enum`, so `release: v2.10.0` is valid
- [x] Document both in CONTRIBUTING (allowed types + Release Process)
- [x] feature-branch-pr skill: require a conventional PR title — its example title would have failed the check
- [x] draft-release-notes skill: note the release PR title convention it depends on
- [x] Verified with commitlint 19.7.1 locally: `release: v2.10.0` and `chore(release): v2.10.0` pass (both failed on the old config), prose titles still fail; workflow YAML parses and both skill frontmatters still load

App health checks are N/A — CI config and docs only.

### DIFF
--- a/.claude/skills/draft-release-notes/SKILL.md
+++ b/.claude/skills/draft-release-notes/SKILL.md
@@ -25,6 +25,11 @@ branch.
 
 If the user names a branch, use it. Don't guess a version — ask if unclear.
 
+The release PR into `main` is titled `Rc vX.Y.Z` (the **Lint PR title** check is
+skipped for `rc-*` head branches) or `release: vX.Y.Z` — see CONTRIBUTING →
+Release Process. Step 2 relies on that `Rc vX.Y.Z` merge-commit title as the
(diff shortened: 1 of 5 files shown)


jashansinghTT (approved):
Was a small but necessary change. LGTM

PR #1250, opened 2026-08-19
feat: add machine-readable NDJSON event stream to the launcher

jashansinghTT:
Gives the launcher a machine-readable mode so a wrapping program (the upcoming desktop launcher) can render bring-up progress natively instead of scraping Rich/ANSI terminal output.

Two entry points:

- `python run.py --json-events` — streams the bring-up as NDJSON on stdout (one JSON object per line): phase lifecycle, progress, warnings, errors with a remediation string, and a final `ready` event carrying the service URLs (app 3000, FastAPI 8001, docker-control 8002) and the detected hardware label. Implies non-interactive: any code path that would prompt emits a `prompt_blocked` event with remediation and exits instead of hanging the consumer. Human output moves to stderr so stdout stays pure NDJSON.
- `python run.py --status --json` — one-shot state dump (per-service health via the existing snapshot helpers, current HEAD, hardware label) through the same emitter, bypassing the TUI.

Every record has the same version-keyed envelope — `{"v":1,"ts":…,"event":…,"phase":…,"detail":{…}}` — documented in `dev-docs/json-events.md`. The events tap the existing phase stepper (`begin_phase`/`end_phase`/notes) and the failure-diagnosis cards; with the flags off every emit is a no-op, so plain interactive runs are unchanged.

- [x] `tt_setup/console/_events.py`: global NDJSON emitter (off by default, stdout/stderr split on enable)
- [x] Lifecycle taps in the stepper + error events (with remediation) from the service/container diagnosis paths
- [x] `--json-events` (Advanced) and `--status --json` (Lifecycle) flags; prompts become `prompt_blocked` + exit 2 in machine mode
- [x] Schema doc in `dev-docs/json-events.md`; options tables updated
- [x] Verified: full launcher suite green (360 tests, incl. new emitter/tap/CLI units and an end-to-end stream test with a PTY ANSI-purity check); `python run.py --status --json` against a live stack emits valid NDJSON with all six services healthy; `--help` shows both flags in the right panels

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -51,6 +51,8 @@ python run.py --purge-all     # wipe containers, volumes, HF-cached models and .
 python run.py --purge-model   # uninstall specific model(s); bare flag opens a picker
 python run.py --logs          # stream all container logs (docker compose logs -f, env-file wired)
 python run.py --info          # re-show the "TT Studio is ready" summary (URLs, mode, hardware)
+python run.py --status --json # one-shot NDJSON state dump (services, HEAD, hardware) — no TUI
+python run.py --json-events   # stream bring-up as NDJSON events (see dev-docs/json-events.md)
 python run.py --no-clear      # start without clearing the terminal; stream full startup detail
 python run.py --report-bug    # bundle logs (logs/tt-studio-logs-ttbr-*.zip) + open a GitHub issue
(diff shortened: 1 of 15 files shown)


jashansinghTT:

PR #1253, opened 2026-08-20
feat: add desktop shell profiles, picker, health gating, and bring-up rendering

jashansinghTT:
Builds the shared desktop shell on top of the Tauri scaffold: connection profiles, a first-run picker (run locally vs connect over SSH), stack health gating, and native rendering of the launcher's `--json-events` NDJSON stream (schema in dev-docs/json-events.md on #1250).

Stacked on #1252; retarget/rebase to `dev` after it merges.

- [x] Connection profile model + CRUD via tauri-plugin-store (`profiles.json` in the app config dir); no secret material in the store
- [x] SSH key passphrases in the OS keychain via `keyring` (secret-service on Linux — `libdbus-1-dev` added to CI); headless machines get a typed `keychain_unavailable` error the UI explains instead of a crash; passphrases are write-only from the webview
- [x] Hardware probe: `/dev/tenstorrent` on Linux (path injectable for tests) decides whether the picker defaults to local or SSH
- [x] Health poller: read-only GETs to frontend `:3000/`, backend `:8000/up/` + `/models/health/`, inference `:8001/health`, docker-control `:8002/api/v1/health`; per-service status pushed as `stack-health` Tauri events; "open the stack" gates on all-green
- [x] Picker + profile editor views with typed IPC wrappers (`src/lib/ipc.ts`)
- [x] Bring-up progress view: phase stepper, notes/warnings, error cards with remediation, prompt_blocked card, ready → navigate transition; NDJSON parser/reducer in `src/lib/events.ts` tolerates bootstrap chatter and unknown event types
- [x] Vitest suite (32 tests: parser/reducer with success, failure, and prompt_blocked fixture streams; picker gating; event rendering) wired into desktop-ci

Verified: `cargo test` (24 tests incl. a stub-HTTP-server health suite), `cargo clippy -D warnings`, `cargo fmt --check`, `npm test`, `npm run type-check`, `npm run build` all pass. The poller was also run against this machine's real ports (stack currently stopped — every service correctly reports `unreachable` and `ready` stays false); an opt-in `--ignored live_stack` test keeps that check runnable.

PR #1254, opened 2026-08-20
feat: add ssh tunnel engine for remote stacks

jashansinghTT:
Milestone D4 of the desktop app: the SSH tunnel engine that lets the launcher front a TT-Studio stack running on a remote machine. Stacked on #1253 (`jashan/desktop-shell`) — retarget to `dev` after that merges.

One russh session multiplexes all forwards as direct-tcpip channels behind local TCP listeners that bind the same port numbers as the remote services (3000, 8000-8002, 8080, 8111, 4000, 7000-7010), since the web app derives its URLs from `window.location.hostname` plus fixed ports. Everything talks to the session through a small `SshTransport` trait, so swapping russh for ssh2 or system-ssh later stays contained to `session.rs`.

- [x] russh session with typed errors per failure mode (dns/refused/timeout/handshake/auth/host key); auth order is ssh-agent (incl. Windows named pipe + Pageant) then key file, with the keychain passphrase from the D2 secrets store
- [x] trust-on-first-use host keys in an app-managed known_hosts file: unknown key surfaces a fingerprint dialog in the launcher; a changed key is a hard fail that is never promptable
- [x] port-forward supervisor: per-listener accept loops, per-tunnel health/last-error, exponential-backoff reconnect (1s doubling to a 30s cap), clean shutdown; fatal errors (auth, host key) stop instead of retrying
- [x] `tunnel-status` events drive a launcher banner; on hard loss while the stack page is showing, the window navigates back to the launcher
- [x] integration tests against a containerized sshd fixture (alpine + busybox httpd): key auth + TOFU accept + HTTP round-trip through the tunnel, reconnect across `docker restart`, changed-key rejection; they skip when docker is unavailable and run on the Linux desktop-ci leg via `--ignored`

Verified locally: cargo test (47 unit tests), the 3 dockerized integration tests, clippy `-D warnings`, rustfmt, `npm test` (38) and `tsc`. Tests bind only OS-assigned high ports, never the real stack's.

### DIFF
--- a/.github/workflows/desktop-ci.yml
+++ b/.github/workflows/desktop-ci.yml
@@ -93,6 +93,13 @@ jobs:
         working-directory: desktop/src-tauri
         run: cargo test
 
+      # SSH tunnel integration tests run against a containerized sshd
+      # (tests/fixtures/sshd) — only the Linux runner has Linux containers.
+      - name: SSH tunnel integration tests
+        if: runner.os == 'Linux'
+        working-directory: desktop/src-tauri
+        run: cargo test --test ssh_tunnel -- --ignored
+
       # Debug + no-bundle keeps this fast and reuses the clippy/test build
       # cache while still exercising the full tauri build path.
       - name: Tauri build (debug, no bundle)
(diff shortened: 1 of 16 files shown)


jashansinghTT:

PR #1252, opened 2026-08-20
feat: add tauri v2 desktop shell scaffold with webview navigation

jashansinghTT:
First cut of a desktop shell for TT-Studio: a new top-level `desktop/` directory with a Tauri v2 app. A small bundled launcher UI (Vite + React + TS + Tailwind, same stack as `app/frontend`) opens in a single window; clicking "Open TT-Studio" invokes an `open_stack(url)` command that validates the URL (http/https on localhost/127.0.0.1 only) and navigates the same window to `http://localhost:3000`, where the real TT-Studio frontend is served. The web frontend is never bundled — it must load from its own origin since it derives API/WebSocket URLs from `window.location`. The remote origin gets no Tauri IPC: capabilities only cover the bundled launcher. The existing web deployment is untouched.

- [x] Tauri v2 scaffold under `desktop/` (launcher UI + `src-tauri/`), pinned stable versions
- [x] SPDX headers on all new sources; `tt_setup/spdx.py` now scans `desktop/` and knows `.rs` files (`python run.py --check-headers` clean for desktop; the one remaining flag is a pre-existing empty `wakeword_control/__init__.py` from #806)
- [x] `open_stack` command with URL validation + unit tests, wired to the launcher button
- [x] `desktop-ci.yml`: 3-OS matrix, path-filtered to `desktop/**` — fmt, clippy `-D warnings`, tests, launcher build, debug `tauri build --no-bundle`
- [x] README documenting the architecture, security model, OS matrix, and a per-OS webview validation checklist

Verified locally on Linux: `npm ci && npm run build`, `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` (4 URL-validation tests), `tauri build --debug --no-bundle`, and a 20s headless launch under Xvfb (window + WebKitGTK init clean). Interactive click-through against a live stack, plus macOS/Windows, are TODO checkboxes in the README. Launcher test suite (325 tests) passes with the spdx.py change.

PR #1255, opened 2026-08-20
feat(desktop): native white-box mode — spawn run.py and stream bring-up

jashansinghTT:
Third slice of the desktop app (D3), stacked on #1253 (\`jashan/desktop-shell\`), which is stacked on #1252 — retarget to \`dev\` after those merge. This makes the QuietBox story real: double-click the app on a Linux box with a Tenstorrent card and it finds (or clones) a tt-studio checkout, runs \`python run.py --no-browser --json-events --no-sudo\`, renders the bring-up natively from the NDJSON stream (dev-docs/json-events.md, #1250), and opens the stack when it's ready. If a stack is already healthy it just attaches — no bring-up spawned.

- [x] Checkout resolution: user-configured path → managed \`~/.local/share/tt-studio/stack\` → shallow-clone of the latest \`v*\` release tag (unit-tested against a local bare-repo fixture, no network in tests)
- [x] Spawner: cwd = checkout root (run.py derives TT_STUDIO_ROOT from cwd), stdout forwarded line-by-line as \`bringup-line\` events, stderr to a rotating log in the app data dir, child survives the venv re-exec as one process
- [x] Attach-first: one health snapshot before spawning anything; picker gets an explicit "Restart stack" action (\`--stop\` then bring-up) that is never triggered automatically
- [x] Prompts can't hang the GUI: \`--no-sudo\` by default, \`prompt_blocked\`/exit-code-2 render a "run one-time setup in a terminal: python run.py" card with a copy button
- [x] Quit dialog: "Keep running" (default — the host services detach and survive the app on purpose) vs "Stop the stack" (runs \`run.py --stop\`, then exits); dialog dismissal maps to keep-running
- [x] SSH fallback polish: the picker now explains *why* local mode is unavailable (non-Linux vs Linux-without-\`/dev/tenstorrent\`)
- [x] Verified: cargo fmt/clippy \`-D warnings\`/test (34 unit + 6 integration), vitest (41), debug tauri build; integration tests drive the spawner against \`tests/fixtures/fake_run.py\` covering success-to-ready, failure-with-remediation, prompt_blocked, and kill/cleanup — no real bring-up was run anywhere
- [x] Attach detection checked read-only against a live local stack (per-service statuses correct)

### DIFF
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -770,6 +770,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "errno"
+version = "0.3.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
+dependencies = [
+ "libc",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


jashansinghTT:

PR #1257, opened 2026-08-20
fix(ui): hide auto-managed fields from the settings dialog

jashansinghTT:
The settings dialog showed read-only, auto-managed entries (JWT secret, TTS API key, and the tt-inference artifact card) alongside the editable secrets. Users can't act on any of them there, so they were just noise and surfaced internals that don't belong in a settings UI.

- [x] Remove the JWT secret, TTS API key, and tt-inference artifact read-only blocks from SettingsDialog; the dialog now shows only the editable Hugging Face token and Tavily API key
- [x] Verified in the running dev stack with a browser click-through of the dialog (only the two editable fields render; save flow untouched)


rnabeelTT (approved):
Looks good

PR #1256, opened 2026-08-20
feat(desktop): ssh remote bring-up and one-click connect flow

jashansinghTT:
Completes SSH-client mode for the desktop shell (D5, stacked on #1254 — review only the last 6 commits until that merges).

One click on a saved machine now does the whole thing: SSH connect → tunnels up → figure out what's on the other side → attach or bring the stack up → the window navigates to the running app. Closing the window while connected asks whether to stop the remote stack or just disconnect.

How the flow decides: once the tunnels report every essential local port bound, the launcher runs three probes over `ssh exec` (`test -f <repo>/run.py`, `python3 --version`, `run.py --status --json`) plus a health poll through the forwarded ports. Healthy → attach; down/partial → stream `run.py --no-browser --json-events` over the exec channel, rendered by the same NDJSON reducer the native path uses (stderr goes to `remote-bringup.log`). A validated repo path is written back to the profile.

Failure cards instead of dead spinners:
- local port 3000 (or another essential port) already taken → explicit card, no silent remapping
- no `run.py` at the repo path → clone instructions + "Edit machine" (auto-clone is out of scope)
- python3 missing or older than 3.12 → clear version card
- probe/tunnel errors reuse the typed SSH error descriptions; cancel tears everything down cleanly

- [x] `ssh exec` helpers: one-shot capture (bounded) + line-streamed exec on the existing russh session, with shell/tilde-safe path quoting
- [x] attach detection: probes + tunnel health folded into a pure, unit-tested classifier
- [x] remote bring-up streamed as `bringup-line` events; cancel closes the session; exit-without-ready synthesizes an error card
- [x] one-click connect flow + progress states + cancel in the launcher
- [x] quit dialog: "Stop the stack on <host>" (streams `run.py --stop` output; failure offers quit-anyway) vs disconnect-only
- [x] e2e: sshd fixture extended with python3 + a fake `run.py`/stub stack; test drives probe → classify(down) → bring-up → ready → HTTP through the forwarded port → stop. Passes locally against Docker (`cargo test --test ssh_tunnel --test ssh_remote_bringup -- --ignored`, 5/5)
- [x] cargo fmt/clippy/test and vitest/tsc green; all test listeners bind high random ports only

PR #1258, opened 2026-08-20
feat(desktop): auto-update — shell updater and stack refresh on launch

jashansinghTT:
Two update layers for the desktop app, both keyed to tagged GitHub releases (the only refs publish-images.yml ships GHCR images for — following raw main would force local image builds).

**Shell updates**: tauri-plugin-updater + tauri-plugin-process, endpoint at the GitHub releases `latest.json`. Silent check on launch (offline never blocks connecting) plus a manual "Check for updates" button with an install-and-relaunch flow. The pubkey in tauri.conf.json is a development placeholder — release CI must generate the real signing key and replace it before shipping updater artifacts; until then the updater fails closed on signature mismatch (documented in desktop/README.md).

**Stack updates**: before every bring-up (native spawn or SSH), `src-tauri/src/update/stack.rs` reads the target checkout's release tag (git locally, `git -C` over ssh exec remotely), compares it to the latest `v*` tag, and when behind runs the stack's own guarded `python run.py --switch <tag>` with streamed output. Policy is a stored setting: auto / ask first (default) / never. It never forces anything: dirty trees and non-release checkouts show "developer checkout detected, skipping stack update" and proceed; offline skips quietly; an already-healthy stack is attached without touching the checkout. No explicit image pull is added — bring-up after a switch pins TT_STUDIO_IMAGE_TAG to the exact tag (tt_setup/env_config/_version.py) and runs `docker compose pull` itself (tt_setup/cli/_run.py "Pull" phase), so the tag's GHCR images arrive as part of normal startup.

Stacking: based on #1256 (jashan/desktop-ssh-bringup, SSH track) with #1255 (jashan/desktop-native-mode, native track) merged in so both spawn paths exist — the first commit is that merge (unified `bringup-exit` payload and combined close-request handling). Review this PR for the seven commits on top.

- [x] tauri updater + process plugins, updater capability, check-on-launch UI
- [x] stack freshness check (pure decision fn + local/ssh gathering)
- [x] `run.py --switch` executor with streamed progress, prompt/auto/never policy
- [x] image-pull coverage verified in tt_setup, documented instead of duplicated
- [x] cargo test (79 lib + 3 updater dry-run + launcher/ssh suites), clippy -D warnings, fmt clean
- [x] vitest 72 passing, tsc + vite build clean
- [x] updater dry-run: real `updater.check()` against a locally served latest.json fixture (newer / older / unreachable feed)
- [x] no real switch or pull was run against a live checkout; tests use temp git repos and fixtures

### DIFF
--- a/desktop/README.md
+++ b/desktop/README.md
@@ -47,6 +47,52 @@ npm run tauri build      # release build + bundles
 Start the TT-Studio stack separately (`python run.py` at the repo root) so
 `http://localhost:3000` is live before clicking "Open TT-Studio".
 
+## Updates
+
+Two layers, both keyed to tagged GitHub releases of tenstorrent/tt-studio.
+Release tags (`v*`) are the source of truth for both: they're the only refs
+`.github/workflows/publish-images.yml` publishes GHCR images for, so
+following raw `main`/`dev` would force local image builds.
+
+### Shell updates (the app itself)
+
+The tauri updater plugin checks the `latest.json` asset on the latest GitHub
+release (endpoint in `src-tauri/tauri.conf.json`). The launcher checks once
+on launch — silently, so being offline never gets in the way of connecting —
+and offers a manual "Check for updates" button whose failures are shown
+(`src/views/UpdateBanner.tsx`). Installing downloads, applies, and relaunches
+via the process plugin.
+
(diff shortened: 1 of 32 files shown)


jashansinghTT:

PR #1259, opened 2026-08-20
ci: build, sign, and publish desktop releases

jashansinghTT:
Desktop milestone D7 — stacked on #1258 (D6, updater); base is `jashan/desktop-updates` and should be retargeted to `dev` once that merges.

One `v*` tag now drives desktop artifacts alongside the GHCR images: `.github/workflows/desktop-release.yml` mirrors publish-images.yml's trigger shape (tag push / release published / workflow_dispatch dry-run with an explicit `publish` input, same GITHUB_TOKEN-pushed-tag caveat) and builds a universal macOS .dmg, a Windows NSIS installer, and Linux .AppImage + .deb (following tt-local-generator's release-deb precedent), uploading them plus the updater `latest.json` to the release via tauri-action.

Every signing layer is skip-if-secret-absent so forks and dry runs stay green: updater artifact signing (config overlay disables `createUpdaterArtifacts` when the key is missing), macOS codesign/notarization (APPLE_* exported only when non-empty), and Windows Authenticode via Azure Trusted Signing (`signCommand` overlay + trusted-signing-cli). Versions are stamped from the tag at build time (`desktop/scripts/stamp-version.mjs`) — git tags stay the source of truth, no version-bump commits.

**Admin action needed before the first real release** (full inventory in `dev-docs/desktop-release.md`):
- `TAURI_SIGNING_PRIVATE_KEY` / `TAURI_SIGNING_PRIVATE_KEY_PASSWORD` — a maintainer runs `npm run tauri signer generate` once, stores the private key in secrets, and replaces the placeholder pubkey in `tauri.conf.json`. Without these, releases ship but in-app updates can't.
- Optional: `APPLE_*` (Developer ID cert + notarization) and `AZURE_*` (Trusted Signing) — unsigned builds still publish, with Gatekeeper/SmartScreen warnings.

- [x] Workflow passes actionlint and yaml parse
- [x] `npm ci && npm run build`, `cargo build` green locally
- [x] Local release `tauri build` produced working `.deb` and `.AppImage` bundles using the same no-key config overlay the workflow uses
- [x] Version stamp script tested against a fake tag (one-line diff in each of tauri.conf.json / package.json / Cargo.toml / Cargo.lock) and rejects non-semver input
- [x] No tags pushed, no workflow dispatches run

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -195,6 +195,32 @@ jobs:
             echo "::warning::APPLE_ID not set — signing but skipping notarization"
           fi
 
+      # Windows Authenticode via Azure Trusted Signing: when the AZURE_*
+      # secrets exist, wire the bundler's signCommand to trusted-signing-cli
+      # (authenticates with the AZURE_TENANT_ID/CLIENT_ID/CLIENT_SECRET env
+      # on the build step). Without them the NSIS installer builds unsigned.
+      - name: Enable Windows code signing
+        if: runner.os == 'Windows'
+        shell: bash
+        env:
+          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+          AZURE_SIGNING_ENDPOINT: ${{ secrets.AZURE_SIGNING_ENDPOINT }}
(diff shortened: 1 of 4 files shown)


jashansinghTT:

PR #1261, opened 2026-08-20
fix: proxy wakeword websocket in nginx prod

jashansinghTT:
The Vite dev server proxies `/ws-api` → backend `/ws` (used by the wakeword voice pipeline, `useWakeWord.ts` → Channels `ws/wakeword/`), but the production nginx config had no matching location, so wake-word WebSockets only worked in dev mode.

- [x] Add a `/ws-api/` location mirroring the Vite rewrite, with the Upgrade/Connection headers and long timeouts a WebSocket needs
- [x] Config syntax validated with `nginx -t` in a throwaway nginx:alpine container (upstream host swapped for a resolvable one)

Not exercised against a live prod frontend container here — the running dev stack was left untouched.


rnabeelTT (approved):
lgtm

PR #1263, opened 2026-08-20
fix(ui): drop the artifact card from the welcome wizard and add a settings icon

jashansinghTT:
The tt-inference artifact card on the token step of the welcome wizard is read-only noise during first-run setup, and the intro step referred to Settings with plain text.

- [x] Remove the tt-inference artifact (read-only) card from the Hugging Face token step
- [x] Show a gear icon in front of "Settings" on the intro step
- [x] Verified by walking both wizard steps in the running dev stack (lint clean; type-check failures are pre-existing in TrainingJobDetailPage)


anirudTT (approved):

anirudTT:
@jashansinghTT good to cherry pick this

PR #1269, opened 2026-08-21
fix: marketplace launch issues

rnabeelTT:
## Marketplace launches: share state across workers, stop racing launches

A user on a QB2 RC box hit `Could not prepare AnythingLLM: 400 Client Error` when launching AnythingLLM. The app had in fact started fine — the error came from redundant launch attempts.

Launch state lived in a module-level dict, but production serves the backend with
`uvicorn --workers 4`. Each worker had its own copy, so the "already launching"
guard only ever saw its own process, and polls answered by other workers reported
`not_installed` — the card flickered, which is what made the user keep clicking.
Both Django caches are `LocMemCache`, so there was no shared store to fall back on.

This also matters now that #1262 has shipped: with `force` finally reaching Docker,
the losing launches would have force-removed the winner's running container
instead of failing with a 400.

### Changes

- Launch jobs move from process memory to a small flock-guarded JSON table under
  the backend volume, shared by every worker of the container. `get_job` /
  `_set_job` / `clear_job` / `is_job_active` keep their signatures, so the views
  are untouched.
- Jobs expire. Each stage refreshes its job every couple of seconds, so one that
  has gone quiet for 60s belonged to a worker that died or a backend that
  restarted; the UI then recovers on its own instead of showing Downloading forever.
- Docker outranks finished jobs. Precedence is now active job -> running container
  -> failed job -> stopped -> not installed, so a persisted job can never pin a
  card that is really running.
- A launch that finds the container already running adopts it rather than
  replacing it, so a racing worker cannot tear down an app that is starting or in
  use. Anything not running is still cleared as a leftover.
- Port allocation and its claim happen in one hold of the table, so concurrent
  launches in different workers cannot pick the same port.
- A readiness timeout now carries the tail of the container's own output into the
  message the card shows — the container is often gone before anyone can run
  `docker logs`, and this is where a permission error or bad mount shows up.

### Verification

Exercised against a live docker-control-service, two OS processes standing in for
two uvicorn workers:

- Worker A claims 3081; worker B sees A's job and is blocked, and allocating for a
  different app returns 3080.
- An abandoned job (last update older than the stale window) is treated as
  inactive and releases its port.
- `_adopt_or_clear` returns adopt for a running container (container survives),
  clears an exited leftover, and no-ops when nothing exists.
- A readiness timeout against a container printing an EACCES line produced
  `... did not respond within 2s. Last output: Error: EACCES: permission denied ...`.
- Every card state re-checked for regressions: not installed, running (with port
  and open path), stopped, pulling with progress bytes, starting, error, guide,
  plus "stale job while container runs" and "old error while container runs" both
  correctly reporting running.
- 6 processes x 60 writes: table stays valid JSON with every writer's last value
  intact. An earlier revision lost writes because the lock was released before the  
  buffer flushed; the flush now happens inside the lock.

### DIFF
--- a/app/backend/model_control/marketplace_utils.py
+++ b/app/backend/model_control/marketplace_utils.py
@@ -9,18 +9,32 @@ gateway by container hostname, so no host-gateway hop is needed. Launching is
 asynchronous — image pulls take minutes — so `start_launch` returns immediately
 and the state of the job it started is reported by `serialize_app`.
 
+Launch state is deliberately not held in process memory: production serves the
+backend with several uvicorn workers, so the worker answering a poll is rarely
+the one running the launch. Everything the UI reports is therefore derived from
+state every worker can see — Docker itself, plus a small locked job file shared
+by the workers of one backend container.
+
 The views for these helpers live in `views.py` (MarketplaceAppsView and friends);
 the app catalog itself is in `shared_config.marketplace_config`.
 """
 
+import fcntl
+import json
 import os
 import threading
 import time
-from typing import Dict, List, Optional, Tuple
+from contextlib import contextmanager
+from pathlib import Path
+from typing import Dict, Iterator, List, Optional, Tuple
 
 import requests
 
-from docker_control.docker_control_client import get_docker_client
+from docker_control.docker_control_client import (
+    ContainerNotFound,
+    get_docker_client,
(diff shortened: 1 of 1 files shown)


jashansinghTT (approved):
lgtm

PR #1274, opened 2026-08-24
feat: pip-installable tt-studio with release auto-update

jashansinghTT:
TT-Studio currently requires cloning the repo to run. This adds a small PyPI package so `pip install tt-studio` (or pipx) gives a `tt-studio` command that installs the latest release into `~/.tt-studio/checkout` and keeps it updated automatically — each launch checks the GitHub releases API and silently moves the managed clone to the newest tag (offline runs just use what's installed).

The shim is stdlib-only and execs the checkout's `run.py`, so every existing flag passes through and the launcher's own bootstrap/GHCR-pull machinery does the rest (a clean tag-detached clone makes `git describe` resolve the release, so prebuilt images are pulled, not built). It never touches a clone it didn't create, skips updates on a dirty tree, and pauses auto-update if the managed checkout was moved onto a branch. `--no-update` and `--pin vX.Y.Z` / `--pin latest` are the escape hatches; `TT_STUDIO_HOME` moves the install root.

- [x] `packaging/pypi/` — shim package (`tt_studio.shim`), no runtime deps
- [x] `publish-pypi.yml` — builds and publishes on every `v*` tag via PyPI trusted publishing (dispatch = dry-run build)
- [x] `tt_setup/shortcut.py` — skip the rc shell-function offer/repair under `TT_STUDIO_MANAGED=1` (it would shadow the console script)
- [x] `tests/test_pip_shim.py` — decision table, tolerant release lookup, flag splitting, pin file (full suite: 346 pass)
- [x] Verified end-to-end from a built wheel: fresh install cloned detached at v2.10.0, simulated stale checkout auto-updated v2.8.0 → v2.10.0, dirty-tree and on-branch guards hold, live API lookup returns v2.10.0
- [x] README quickstart + `dev-docs/pip-package.md`

Admin prerequisite before the first release publish: create the `tt-studio` project on pypi.org and register this repo + `publish-pypi.yml` as a trusted publisher (plus a `pypi` environment in repo settings).

## How to test (before the PyPI publish)

The package isn't on PyPI yet, so install the shim from this branch's `packaging/pypi/`. Note the shim doesn't run this branch's code — it clones the **latest GitHub release tag** into `~/.tt-studio/checkout` and execs that checkout's `run.py`.

1. Stop any running dev stack first (it holds ports 3000/8000/8080/8111): `cd ~/tt-studio && python run.py --stop`
2. Install the shim in an isolated env — pipx is the cleanest "brand new terminal" simulation:
   ```bash
   pipx install ./packaging/pypi
   # or the wheel route (closer to what PyPI ships):
   python -m pip install build && python -m build packaging/pypi
   pipx install packaging/pypi/dist/tt_studio-*.whl
   ```
3. Run `tt-studio` from anywhere outside the repo. Expected: clone at `~/.tt-studio/checkout` detached at the latest release tag, one shim line, then the normal launcher pulling **prebuilt GHCR images** (not building), since the clean tag-detached clone resolves via `git describe`.
4. Verify: `--shim-version`, `--no-update`, `--pin v2.8.0` → `--pin latest` (auto-update path); passthrough flags (`--info`, `--stop`); no shell-shortcut offer (suppressed under `TT_STUDIO_MANAGED=1`); your dev clone untouched (no `.tt-studio-managed` marker); fresh `.env` in the managed checkout (needs `HF_TOKEN` again for gated models).
5. Cleanup: `tt-studio --stop` (or `--purge-all`), `pipx uninstall tt-studio`, `rm -rf ~/.tt-studio`.


packaging/pypi/pyproject.toml
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+[project]
+name = "tt-studio"
+# Placeholder — .github/workflows/publish-pypi.yml injects the release tag
+# (vX.Y.Z -> X.Y.Z) before building, so the PyPI version always matches the
+# GitHub release the shim will install.
+version = "0.0.0"
```
knauth, line 9: Should be something like "__DEV_TEST_BUILD__" so there is no confusion that this is a valid semver.

knauth (changes requested):
This is a usable way of doing it, but I would vastly prefer that the entire project live in the installed package, rather than this shim technique. Can you clarify why that decision was made and whether we can do it differently?

PR #1275, opened 2026-08-24
feat: file bug reports as Jira tickets with log bundle attached

jashansinghTT:
Bug reports from both the launcher (`python run.py --report-bug` and the crash prompt) and the web UI's Report Bug dialog now file a Jira ticket in DEVSTACK instead of a GitHub issue: the diagnostics ZIP is attached to the ticket automatically, the failing component is classified from the traceback/logs, and the ticket is assigned to that component's primary codeowner (mapping in `app/backend/shared_config/component_owners.json`, mirroring CODEOWNERS). Without `JIRA_EMAIL`/`JIRA_API_TOKEN` in `.env`, both paths fall back to the existing pre-filled GitHub issue URL, so community setups keep working.

- [x] `tt_setup/jira_report.py` — CLI Jira client (create + attach + assignee lookup), never raises inside the crash handler
- [x] `logs_control/jira_client.py` + `JiraIssueView` (`/logs-api/jira-issue/`) — backend twin, builds the ZIP server-side and attaches it; `github-issue/` kept for back-compat
- [x] Bug Report modal now creates the ticket directly; manual ZIP-attach instructions only shown on the GitHub fallback
- [x] New env vars documented in `.env.default` and `--help-env`; passed to the backend container in docker-compose
- [x] Verified: 348 launcher tests + 10 backend tests green; fallback path exercised live against the dev stack (`/logs/jira-issue/` and via the frontend proxy); full create+attach flow verified end-to-end against a mock Jira server from both the CLI and inside the backend container (correct project/issuetype/labels/assignee, ZIP uploaded with `X-Atlassian-Token: no-check`)

Not yet verified against real Jira: no API token was available in this environment, so the first real ticket should be created by someone with a token (checks DEVSTACK required fields and confirms the teammate emails in `component_owners.json` resolve — a wrong email just leaves the ticket unassigned).


jashansinghTT:
Closing in favor of #1292 — bug reports now go through the support@tenstorrent.com inbox (which already files Jira tickets from email), so the direct Jira REST integration and its API-token setup are no longer needed.

PR #1278, opened 2026-08-24
fix: stop losing deploy secrets when deploys run concurrently

anirudTT:
Deploys submitted while another deploy was in flight lost their `JWT_SECRET` and `HF_TOKEN` and failed instantly. The Voice Agent page hit this every time because it fires three deploys at once, while retrying a single model always worked.

Secret resolution was conditioned on the shared `os.environ`, which another in-flight job sets and then wipes on exit. Each job now carries its own copy.

- [x] Resolve secrets per job in `_resolve_secret_env_vars()`, independent of `os.environ`
- [x] Same fix for `HF_TOKEN`
- [x] Show the backend's real error on the card instead of a generic string
- [x] Regression tests, verified failing against the old code


rnabeelTT (approved):
LGTM

PR #1279, opened 2026-08-24
fix: stop Docker Control supervisors accumulating and fighting over port 8002

anirudTT:
Every `python run.py` left another Docker Control supervisor behind and they fought over port 8002 — five were running on one machine, with restart counters in the thousands. Image pull progress lives in that service's memory, so each change of ownership dropped in-flight pulls and surfaced in the UI as "Deployment failed".

The "service already running" check ran after the code that frees the port, so it could never fire.

- [x] Adopt a healthy service instead of replacing it
- [x] Stop the previous supervisor before freeing the port, so nothing respawns a competitor
- [x] Bound the restart loop so a permanent fault stops with a message pointing at port 8002
- [x] Regression tests; full suite green


jashansinghTT (approved):
LGTM, deployment timed out worked at the present

PR #1281, opened 2026-08-25
fix: keep image-pull progress across backend workers and restarts

anirudTT:
Deploying a model whose image isn't cached would show "Deployment failed" on the card while the pull was actually running fine.

The backend runs uvicorn with `--workers 4`, but image-pull progress lived in a module-level dict. The worker that minted a `pull_id` and ran its pull thread was usually not the worker serving the client's progress polls, so roughly 3 in 4 polls landed on a process that had never heard of the job and returned a bare `not_found`. The Voice Agent card treated the first `not_found` as fatal — no grace period, unlike the two shared hooks — so a healthy deploy died within seconds with no explanation. This reproduces in prod mode; dev mode is single-worker, which is why it looked intermittent.

State now lives in a small JSON store in the persistent volume, one file per pull with `flock` around read-modify-write, following the `deployment_store.py` pattern already used for deployments. If the volume isn't writable it falls back to per-process memory, so the degraded case is no worse than today rather than making every pull look orphaned.

The rest is about reporting an interrupted pull honestly, and using telemetry the UI already receives but discarded.

- [x] Durable cross-process pull store — fixes the `--workers 4` sharding
- [x] Detect a pull whose owner died instead of reporting "pulling" forever
- [x] Stall detection when the pull backend stops answering, instead of looping the full 2h at 0%
- [x] Reconcile an unknown pull against reality before calling it a failure, and release the chip slot its placeholder was holding
- [x] Give the orphan response a message — it was the only terminal branch in `DeploymentProgressView` without one
- [x] 90s `not_found` grace in the Voice Agent poller, matching `useDeploymentProgress` and `useActiveDeployments`
- [x] Attach to an in-flight pull of the same image rather than starting a second one — speech-to-text and text-to-speech share `tt-media-inference-server` and were pulling it twice
- [x] Show the bytes/speed/layers the backend already sends, and withhold the ETA until it stabilises (a pull that finished in 16 min was projecting 106 min at 5%)
- [x] Refetch device occupancy on focus and after deploys, so the Deploy button can't stay disabled against stale state and silently swallow clicks
- [x] "View logs" resolves `imgpull_` ids instead of 404ing
- [x] 12 new tests, including cross-process visibility and concurrent writers
- [x] Verified against a running backend: an entry written by a separate process is served correctly through the HTTP API (49%, bytes, speed, layers) where the old code returned `not_found`; orphaned and stalled pulls both return actionable messages
- [x] tsc + eslint clean; ruff unchanged from baseline (11 pre-existing findings before and after)

One thing I deliberately did not change: `image_exists` failing open to an inline deploy. That is `image_pull`'s stated design rule — never block a deploy — so I only made the log say what the user loses (the progress bar).

### DIFF
--- a/app/backend/docker_control/image_pull.py
+++ b/app/backend/docker_control/image_pull.py
@@ -30,20 +30,24 @@ from __future__ import annotations
 
 import threading
 import time
-from typing import Callable, Dict, Optional, Tuple
+from typing import Callable, Optional, Tuple
 
 from shared_config.logger_config import get_logger
 from docker_control.docker_control_client import get_docker_client
+from docker_control import image_pull_store as store
 
 logger = get_logger(__name__)
 
-# pull_id -> snapshot dict (see _new_entry)
-_image_pull_jobs: Dict[str, dict] = {}
-_lock = threading.Lock()
+# Progress lives in image_pull_store (JSON in the persistent volume), NOT in a
+# module global: the backend runs multiple uvicorn workers, so the process that
+# mints a pull_id is usually not the one serving the client's progress polls.
 
 _POLL_INTERVAL_SECONDS = 1.5
 _PULL_TIMEOUT_SECONDS = 2 * 60 * 60  # 2h hard cap; large multi-GB images
-_ENTRY_TTL_SECONDS = 3600            # evict finished entries after an hour
+# How long the pull backend may stop answering before we call the pull dead.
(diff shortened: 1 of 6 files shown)


jashansinghTT (approved):
Survived restart and timed out when necessary. Hence this PR looks good to me at the present.

PR #1282, opened 2026-08-26
feat: transcribe uploaded audio files in speech to text

stisiTT:
Supersedes #1249, which was closed before this had been run against a real model. Same branch, now verified end to end on hardware.

Speech to Text could only transcribe live microphone input. This adds a drag and drop upload panel so an existing audio file can be transcribed, aimed at 40 to 60 minute recordings.

The reason this is more than a file picker: the media server caps a request at 60 seconds of audio, and when you exceed it, it does not return an error. `audio_manager._validate_and_truncate_duration` slices the array, logs a warning inside the container, and returns 200 OK. Upload a 30 minute file in one piece and you get a confident transcript of the first minute with nothing on the client saying the other 29 were discarded. So the file is split client side, at the quietest point near each boundary to avoid cutting words in half, and transcribed one chunk at a time.

- [x] Drag and drop upload panel, reusing the existing `FileUpload` primitive (adds an optional `accept` prop to it)
- [x] Decode once, split at silence, transcribe sequentially with per segment progress and a cancel button
- [x] Retry a failed segment, keep the rest, and persist a partial transcript so a refresh or cancel does not discard a long run
- [x] Render the transcript as timestamped segments
- [x] Raise the transcription proxy timeout from 5s to a 10s connect / 900s read pair, and return real errors: 504 on timeout, 502 when unreachable, upstream body instead of a bodyless 500
- [x] Raise the client abort from 30s, which would have killed a long run
- [x] Add `client_max_body_size` to `/models-api/`, which was on nginx's 1 MB default in production
- [x] Downmix to mono instead of interleaving stereo, halving the payload for non microphone sources
- [x] Replace `alert()` with the standard toast, and stop rebuilding blob URLs on every render
- [x] Delete a byte identical duplicate definition of both speech recognition views (Python kept the second, so edits to the first were silent no ops) and two log lines that wrote the cloud auth token in plaintext

Verified on a P300x2 box against `whisper-large-v3`, transcribing a 30 MB / ~32 minute mp3:

- 77 chunks, 77 x 200 OK, no retries and no timeouts
- zero truncation warnings in the model container, so no audio was silently dropped
- 2 minutes 12 seconds wall clock, about 15x realtime
- CI on the previous PR passed the full suite on this commit: prod image build and smoke test, dev image build, production build, Playwright render tests, SPDX and Cycode

Two things a reviewer should know, neither of which affected the run above.

`prompt` is accepted by the API and assigned in `whisper_runner.py:188`, but `GenerationParams` has no `prompt` field and the only `generator.generate()` call site never passes one, so the server discards it. I confirmed that empirically: three different prompts produce byte identical output. `PROMPT_CARRY_OVER` in `apiClient.tsx` is therefore currently inert and should be flipped to `false` until it is fixed upstream, along with `language` and `task` which are dropped the same way.

The comment on the `model_config.py` change describes the server dropping to 3 second chunks on a multi worker box. That is what the tt-media-server README documents, but on the deployed 0.15.0 image `_calculate_audio_chunk_duration` returns 15 unconditionally. The pin is still worth having, the stated reason needs correcting.

Also worth noting the Voice Agent keeps byte identical copies of `apiClient` and `waveConverter`, so none of these fixes reach it. Left alone here rather than widening the diff.


jashansinghTT (approved):
Image

Was able to transcribe a file hence this looks good to me in this case

PR #1285, opened 2026-08-26
fix: remove Z-Image-Turbo from the model catalog

jashansinghTT:
Z-Image-Turbo deploys wedge Blackhole P300 devices during warmup — one attempt died at device init (eth-core 24-25 timeout), the next hard-hung the board mid kernel-compile badly enough that `tt-smi` timed out on the host and the chips needed a board reset. Until the media image ships a tt-metal/firmware combo validated on fw 19.7.0, the model shouldn't be offered.

- [x] Drop the Z-Image-Turbo entry from `models_from_inference_server.json`
- [x] Add `EXCLUDED_MODEL_NAMES` to the sync script so a resync can't reintroduce it (verified: resync against the 0.20 artifact prints "Excluded 1 model(s): Z-Image-Turbo" and the entry stays gone)
- [x] `test_sync_models.py` — 30 passed

### DIFF
--- a/app/backend/shared_config/models_from_inference_server.json
+++ b/app/backend/shared_config/models_from_inference_server.json
@@ -3,7 +3,7 @@
     "artifact_version": "0.20.0",
     "generated_at": "2026-08-17T18:22:32.785689+00:00"
   },
-  "total_models": 60,
+  "total_models": 59,
   "models": [
     {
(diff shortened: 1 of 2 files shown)


rnabeelTT (approved):
lgtm

PR #1286, opened 2026-08-26
fix: improve error logging and UI messaging for missing HF models

### LINKED ISSUE #803: Improve Error Logging & UI Messaging for Missing Hugging Face Models Description
## Description

Currently, when a requested model is not available on Hugging Face through the configured environment variable, the system fails with a generic:

- `500 Internal Server Error`
- `"died unexpectedly"`

This creates confusion for users because the request itself may be valid, but the underlying model configuration is unavailable or incorrectly set.

Instead of exposing an unclear backend failure, we should provide:

1. Better backend logging for debugging
2. User-friendly UI error messaging
3. Proper distinction between:
   - Invalid user request
   - Missing/unavailable model configuration
   - Runtime/backend failures

---

## Current Behavior

- User sends a valid request
- Backend attempts to load a model from Hugging Face
- If the model is unavailable/misconfigured through the environment variable:
  - Request crashes
  - Generic `500` error is returned
  - UI shows `"died unexpectedly"`

This makes it difficult for users to understand what actually went wrong.

---

## Expected Behavior

If the configured model is unavailable or missing:

- Backend should log the exact failure reason
- API should return a meaningful structured error
- UI should display a user-friendly message such as:

> "The requested model is currently unavailable or not configured properly."

or

> "The model could not be loaded from Hugging Face. Please try again later or contact support."

The UI should avoid exposing raw internal server errors where possible.

---

### Backend

- Add explicit exception handling for:
  - Missing Hugging Face model
  - Invalid environment variable
  - Authentication/configuration issues
  - Model loading failures

- Improve structured logging with:
  - Requested model name
  - Environment variable value
  - Hugging Face error response
  - Stack trace (internal only)

- Return appropriate status codes/messages instead of generic `500`

Example response:

```json
{
  "error": "MODEL_UNAVAILABLE",
  "message": "The configured Hugging Face model could not be loaded."
}

AnkittChauhan:
### Summary
When a model repo was missing or mistyped on Hugging Face (404), TT Studio didn't catch it during pre-check and would still proceed to pull the full Docker image. Once the container crashed on startup trying to download weights, the backend labeled it as "unknown", which made the UI say "Died Unexpectedly" and suggest a hardware reset.

- Fixes #803

### Changes
- **DeployView**: Catch 404s in the Hugging Face pre-check and reject early with `400` (`hf_model_not_found`) so we don't pull images or reserve chip slots on bad repos.
- **deployment_sync**: Recognize `RepositoryNotFoundError`, `EntryNotFoundError`, and 404 logs so they're classified as `hf_model_not_found` rather than `unknown`.
- **UI**:
  - Show an error toast on deployment if the repo isn't found.
  - Added a "Model Not Found on Hugging Face" badge in Deployment History.
  - Replaced the board reset warning banner in the log dialog with a specific Hugging Face error message.
- **Tests**: Added test cases in `docker_control/tests.py` for 404 pre-checks and classification.

### Verification
- Tested unit tests in `docker_control/tests.py`.
- Tested deploying a non-existent HF model repo: request is rejected immediately with a toast without pulling container images.


jashansinghTT (approved):
LGTM, thanks for the better UI messaging for the missing HF tokens.

PR #1288, opened 2026-08-26
docs: refresh the tt_setup module map and cross-link the CLI skills

anirudTT:
The `tt-studio-cli` skill's file map had fallen behind `tt_setup/` — it was missing `switch.py`, `image_source.py` and `config_store.py`. That matters because the skill tells readers to "treat this skill's table as the current layout," so the stale row was actively misleading.

Also links the two CLI skills to each other. They cover complementary ground — `cli-design` is the repo-neutral design language, `tt-studio-cli` is the mechanics for this repo — but neither mentioned the other, so a reader of one never learned the other existed.

- [x] Add the three missing modules to the standalone-modules row
- [x] Note in `tt-studio-cli` that it's TT-Studio-specific, and point at `cli-design` for other repos
- [x] Point `cli-design` at `tt-studio-cli` for the concrete mechanics
- [x] Verified: frontmatter parses and `name` matches each directory; every module named in the map exists in `tt_setup/`, and no top-level `tt_setup` module is omitted (16/16 in sync); both cross-referenced skill directories exist

Docs-only, so app endpoint checks are N/A — no launcher code changed.

Both skills were also packaged for the DX team marketplace (tenstorrent/dx-team#3); this keeps that copy and this one from diverging.

### DIFF
--- a/.claude/skills/cli-design/SKILL.md
+++ b/.claude/skills/cli-design/SKILL.md
@@ -47,7 +47,9 @@ python3 reference/cli_skeleton.py --help   # the command surface  (then: no flag
 | `reference/test_cli_output.py` | copy into `tests/`: pure helpers, stream parsing, PTY + non-TTY render checks |
 
 Reference implementation in the wild: TT-Studio's `tt_setup/` (+ its house-rules doc
-`dev-docs/launcher-terminal-design.md`).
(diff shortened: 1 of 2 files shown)


jashansinghTT (approved):
LGTM

PR #1289, opened 2026-08-26
fix: don't block ungated model deploys when no HF token is saved

jashansinghTT:
With no HF token saved, the deploy step's access check marked every model's repo as blocked without ever asking Hugging Face, so public models like Qwen3.5-9B showed "Hugging Face Access Required" and couldn't be deployed. Public repos answer 200 to anonymous requests, so a token isn't needed for them.

- [x] Check explicitly requested repos anonymously when no token is saved; public repos read granted, gated ones read no_token (an anonymous 401 is a missing token, not a bad credential)
- [x] Welcome-flow default check (all-gated set) keeps its no-network short-circuit
- [x] Added HfCheckView tests; full `manage.py test api` passes
- [x] Verified against the live dev stack: `/settings/hf-check/` now returns granted for Qwen/Qwen3.5-9B and no_token for meta-llama/Llama-3.1-8B-Instruct with no token saved

### DIFF
--- a/app/backend/api/hf_access.py
+++ b/app/backend/api/hf_access.py
@@ -49,6 +49,11 @@ def check_hf_repos(token: str, repos: List[str]) -> List[Dict]:
 
     Mirrors DeployView's gate, including its model_index.json retry: diffusers
     repos have no root config.json, so a gated one would read as "error".
+
+    Works without a token too: public repos answer 200 anonymously, so an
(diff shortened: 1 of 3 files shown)


anirudTT (approved):

anirudTT:
good to cherry pick this into the open rc

PR #1299, opened 2026-09-01
feat: drive external image models over the API they actually serve

rnabeelTT:
## Summary
Externally registered image models came up unusable — labelled as chat models, with a service route that 404s. tt-metal's DiT image servers (models.tt_dit) serve `POST /generate` plus `/jobs/<id>` polling rather than `/v1/images/generations`, and our name-based type guess falls back to chat when it recognises nothing.

Registration now reads the container's OpenAPI document to derive both the model type and the route it actually serves, which beats guessing from weight names. The two hardcoded branches in the image view become a dialect table, so a new
serving stack is a new row instead of another branch.

## Changes
- Derive model type and service route from the container's OpenAPI document
- Persist the observed route so it survives a backend restart
- Replace the hardcoded image branches with a dialect table (openai / media / tt_dit)
- Forward diffusion params (steps, guidance, seed, size) where the server accepts them
- Bound the job poll so a wedged generation can't hang the request
- Verified against a running FLUX.2-dev container: 202 + job_id → done → valid PNG
- 30 backend tests passing (14 new)

### DIFF
--- a/app/backend/docker_control/deployment_store.py
+++ b/app/backend/docker_control/deployment_store.py
@@ -207,6 +207,7 @@ class _Manager:
                 "jwt_secret": kwargs.get("jwt_secret", None),
                 "model_type": kwargs.get("model_type", None),
                 "hf_model_id": kwargs.get("hf_model_id", None),
+                "service_route": kwargs.get("service_route", None),
             }
             data["next_id"] += 1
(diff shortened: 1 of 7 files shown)


jashansinghTT (approved):
LGTM

PR #1301, opened 2026-09-01
feat: single-instance guard and backend host port auto-increment

jashansinghTT:
Starting TT Studio used to hard-fail when something else was listening on port 8000, and booting while another checkout's stack was running would silently take over (or collide with) its containers. This PR makes both cases explicit and safe.

- [x] Single-instance guard: startup detects a running TT Studio started from a *different* checkout (via the compose working-dir labels on its containers) and either stops it after an explicit confirmation, or exits with instructions. Two stacks can't coexist anyway — they share container names, host ports, and the Docker network.
- [x] Backend host port mapping is now `${BACKEND_PORT:-8000}:8000`; when a non-TT-Studio process holds the port, run.py bumps to the next free one (skipping ports reserved for the other services) instead of failing. The container-internal port stays 8000, so in-network consumers (vite proxy, nginx, agent) are unaffected.
- [x] A port held by our *own* backend container is kept, so restarts stay on 8000. The guard runs before port resolution, so by then any TT Studio holder is ours.
- [x] The chosen port is persisted to `.env` and announced; health waits, `--status`, and the startup port check read BACKEND_PORT instead of hardcoding 8000.
- [x] Verified with unit tests (full `tests/` suite green, 16 new cases), a real-socket check of the port resolver, `docker compose config` rendering (8000 default / 8005 with BACKEND_PORT=8005), and live read-only detection of an actual second-checkout stack running on this box. Not exercised against a full stack bring-up.


rnabeelTT (approved):
LGTM

PR #1305, opened 2026-09-02
feat: enable compatible p300x2 models on p150x4

rnabeelTT:
## Summary
P150x4 and P300x2 are the same four-chip Blackhole mesh for vLLM, so chat models that only ship a p300x2 spec can now deploy on P150x4 (and the other way around). Media models stay on the board's own spec.

- Fall back to the other four-chip vLLM spec when this board has none, and surface those models in the catalog
- Pin FLUX/Wan images and raise trace_region_size on P150x4 so traced warmup does not TT_FATAL
- Mirror media trace_region overrides into TRACE_REGION_SIZE for the container
- Skip editor-skill symlinks when extracting inference-server artifacts (fixes Ubuntu 22.04 / Python 3.10.12)
- Give the CI catalog fetch a 180s timeout so it does not race a full Docker probe


jashansinghTT (approved):
LGTM

PR #1306, opened 2026-09-03
fix(chat): stream thinking process live for reasoning models

### LINKED ISSUE #1298: Thinking process is not streamed for reasoning models (e.g. diffusiongemma-26B-A4B-it)
## Description

For models that emit a thinking/reasoning block (e.g. `diffusiongemma-26B-A4B-it`), the chat UI does not stream the thinking process as it is generated. Instead, nothing is rendered until the model finishes, and then the thinking block and the final answer appear together all at once.

## Expected behavior

- The thinking block should stream token-by-token while the model is reasoning, so the user sees progress immediately.
- Once reasoning ends, the final answer should continue streaming below it (matching the existing collapsible "Show/Hide thinking process" UI).

## Actual behavior

- The UI stays empty during the entire reasoning phase, then jumps straight to the completed chat output with the thinking block already fully populated.
- This makes TTFT feel very long — observed TTFT of ~24s and ~14s on diffusiongemma while the screen showed nothing.

## Steps to reproduce

1. Deploy `diffusiongemma-26B-A4B-it` and open its chat page.
2. Send any prompt (e.g. "Write me a poem.").
3. Observe that no thinking tokens stream in; the whole response (thinking + answer) renders only after generation completes.

## Notes

- Likely related to how the streaming handler parses/buffers the thinking block (recent work in `fix chat ui thinking block for diffusion gemma`, f6c553e) — the block may only be emitted to the UI after the closing tag is seen rather than incrementally.

AnkittChauhan:
## Description

Reasoning models like `diffusiongemma-26B-A4B-it` weren't streaming their thinking process in real time—the chat UI stayed blank while reasoning, and then rendered the thinking block and response all at once after generation finished.
 - Fixes #1298
## What changed
- **StreamingMessage.tsx**:
  - Sliced live thinking text by index instead of using an anchored `^<think>` regex (which failed whenever leading whitespace or newlines were present).
  - Derived `isThinkingActive` synchronously on render to eliminate the 1-tick `useEffect` delay.
  - Normalized `<thought>` and `<reasoning>` tags to support Gemma-style thinking blocks.
- **runInference.ts**:
  - Stripped variant thinking tags from `reasoning_content` deltas.
- **model_utils.py**:
  - Switched from `aiter_text()` to `aiter_lines()` in the SSE streaming proxy so chunk boundaries don't drop tokens or fail JSON parsing.

## How it was tested
- Since I don't have physical TT hardware, I mocked the inference server and SSE stream to test and validate the end-to-end flow locally in the browser.
- Verified that reasoning tokens stream incrementally inside the thinking box.
- Verified that the thinking panel collapses to "Show thinking process" and the answer streams smoothly underneath once reasoning finishes.
- Confirmed `npm run type-check` and `npm run build` pass cleanly.

## Working

https://github.com/user-attachments/assets/0fd1377c-a08d-4c8d-8c39-c0467282e219

### DIFF
--- a/app/backend/model_control/model_utils.py
+++ b/app/backend/model_control/model_utils.py
@@ -440,46 +440,52 @@ async def stream_response_from_external_api(url: str, json_data: dict, auth_toke
             if te != "chunked":
                 logger.warning(f"Unexpected transfer-encoding from vLLM: {te!r}")
 
-            async for chunk in response.aiter_text():
-                logger.debug(f"stream_response_from_external_api chunk:={chunk}")
-                if chunk.startswith("data: "):
-                    new_chunk = chunk[len("data: "):].strip()
+            async for line in response.aiter_lines():
+                line = line.strip()
+                if not line:
+                    continue
+                logger.debug(f"stream_response_from_external_api line:={line}")
+                if line.startswith("data: "):
+                    new_chunk = line[len("data: "):].strip()
 
                     if new_chunk == "[DONE]":
-                        yield chunk
+                        yield "data: [DONE]\n\n"
                         stats = tracker.get_stats()
                         logger.info(f"ttft and tpot stats: {stats}")
                         yield "data: " + json.dumps(stats) + "\n\n"
                         break
 
                     elif new_chunk != "":
-                        chunk_dict = json.loads(new_chunk)
+                        try:
+                            chunk_dict = json.loads(new_chunk)
 
-                        # Track TTFT/TPOT from content delta chunks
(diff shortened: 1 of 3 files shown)


jashansinghTT (approved):
@AnkittChauhan I tested this PR and it works flawlessly. We are beyond thankful for your work, Ankit. Just keep this up! Thank you so much for making these contributions.

AnkittChauhan:
Thank you @jashansinghTT  that means a lot!

PR #1314, opened 2026-09-07
fix: prevent orphaned supervisor loops on stop or restart

### LINKED ISSUE #1307: Docker Control Service (port 8002) leaves orphaned supervisor loops that survive --stop and restarts
## Summary

The Docker Control Service on port 8002 is not reliably stopped or restarted by the launcher. Over time a box ends up with **more than one supervisor loop** for the service. Only one of them can hold the port; the others sit in the background retrying the bind every 3 seconds, forever. `python run.py --stop` does not remove them, a fresh `python run.py` does not remove them, and they survive reboots of the stack until someone hunts them down by hand.

This is the root cause behind a cluster of symptoms we have been working around for a while:

- "port 8002 is already in use" warnings on every startup even right after a `--stop`
- the Docker Control log being unreadable because two processes are appending to the same file at the same time
- the service log filling with thousands of `address already in use` restart lines
- `--stop` followed by `run.py` racing against a ghost process for the port, so which instance actually ends up serving the backend is luck

## What is actually happening

`start_docker_control_service()` in `tt_setup/services/_docker_control.py` launches the service through a small bash wrapper script (written to `/tmp`, detached with `start_new_session=True`). The wrapper runs uvicorn in a `while true` restart loop so the service comes back on its own if it crashes. The wrapper writes **its own PID** to `logs/docker-control-service.pid`, then spawns uvicorn as a child.

Everything that tries to tear the service down looks at only two things:

1. the single PID in the PID file
2. whatever process currently **holds the socket** on port 8002 (via `lsof` / `ss`)

Neither of those ever points at a wrapper whose child is not currently bound. So the moment a second wrapper exists, the first one becomes invisible to the launcher.

How you get a second wrapper (all of these happen in normal use):

- **Plain restart.** A normal `python run.py` does not go through `cleanup_docker_control_service()`. It only calls `kill_process_on_port(8002)`, which kills the **uvicorn child**, not the wrapper. The wrapper's restart loop respawns uvicorn 3 seconds later. Meanwhile the new run has started its own wrapper and **overwritten the PID file**. You now have two supervisors and one PID file, and the old one can never be found again.
- **Failed start.** If the health check times out, `_stop_supervisor()` terminates the wrapper that *this* run started, but the previously orphaned one keeps going.
- **`--purge-all` / venv wipe while the service is up.** The orphan's loop keeps trying `.venv/bin/uvicorn` which no longer exists, logs `exit code 127`, sleeps 3s, tries again. Once the venv is recreated it happily comes back and starts fighting for the port.
- **Dev mode (`--reload`).** The reloader parent and the actual server are different processes. Killing the port holder kills the server; the reloader respawns it. The wrapper already has a comment acknowledging this exact problem.

Once two wrappers exist, `--stop` kills the one in the PID file plus whatever is on the port right now, and the ghost immediately wins the bind. The next `run.py` sees 8002 in use, warns, kills the ghost's *child*, and starts its own uvicorn. Whichever binds first wins; the loser goes back to looping. Both wrappers append to the same `docker-control-service.log`, which is why the log looks interleaved and useless.

## Evidence from a real box (today)

```
$ ps -eo pid,ppid,lstart,cmd | grep -E "tmp[a-z0-9_]*\.sh|uvicorn api:app"
 260809       1  Sep  4 01:09:23  /bin/bash /tmp/tmpera45p4b.sh .../docker-control-service .../docker-control-service.pid .venv .../docker-control-service.log
1329061       1  Sep  4 16:23:56  /bin/bash /tmp/tmp022704v1.sh .../docker-control-service .../docker-control-service.pid .venv .../docker-control-service.log
1329063 1329061  Sep  4 16:23:56  .venv/bin/python3.12 .venv/bin/uvicorn api:app --host 0.0.0.0 --port 8002

$ cat logs/docker-control-service.pid
1329061
```

Two wrappers. The PID file only knows about the one from 16:23. The one from 01:09 has been orphaned for ~15 hours and survived several `--stop` / restart cycles plus a venv wipe.

Its restart counter in the log:

```
[Fri Sep  4 03:56:43 PM UTC 2026] Docker Control Service exited with code 127 (restart #16293) — restarting in 3s...
/tmp/tmpera45p4b.sh: line 38: .venv/bin/uvicorn: No such file or directory
...
ERROR:    [Errno 98] error while attempting to bind on address ('0.0.0.0', 8002): address already in use
[Fri Sep  4 04:33:09 PM UTC 2026] Docker Control Service exited with code 1 (restart #16940) — restarting in 3s...
```

Nearly 17,000 restarts in one day from a single ghost. The wrapper script it was launched from was deleted from `/tmp` right after start (by design), so there is nothing on disk to correlate it back to a TT Studio run other than the command line.

The 16:22 archived log is ~940 KB of almost nothing but this loop.

## Why this matters beyond the noise

- The backend only sees Docker through this service. If the ghost wins the port with a stale venv or stale code, the UI's Models Deployed view is talking to whatever old version happens to be bound, not what the current checkout expects.
- Every `run.py` start currently pays the "free port 8002" path (kill + wait + retry), and shows a warning the user cannot act on.
- On shared boxes and CI runners these ghosts accumulate across users and sessions. The 8002 port being taken by someone else's ghost is a real failure mode for the runner.
- `--stop` is supposed to mean stopped. Right now it means "stopped the one we remember".

## Suggested direction (not prescriptive)

A few things that would each help; the first two are probably the minimum:

1. **Track the supervisor by something that survives PID-file overwrites.** Either refuse to start a second wrapper while one is alive (check the PID file *before* truncating it and stop the old wrapper properly), or record every wrapper we launch (e.g. a PID directory, or a stable process-group / session id we can `pkill -s`) and have cleanup walk all of them.
2. **Kill the wrapper, not the child, when freeing the port.** `kill_process_on_port(8002)` should walk up to the parent bash wrapper (or send to the process group / session) so the restart loop cannot resurrect anything. Same fix likely applies to the inference API wrapper on 8001, which uses the same pattern.
3. **Make the normal `run.py` start path go through the same cleanup as `--stop`** for host services, so a restart is a real restart.
4. **Give the wrapper a way to notice it is a ghost.** E.g. bail out after N consecutive `address already in use` exits, or check that the PID file still contains its own PID before each respawn and exit if it does not.
5. **Cap or back off the restart loop.** 3 seconds forever with no ceiling is what turns one orphan into a 1 MB log per day.

## Reproduction

1. `python run.py` (full stack up, service healthy on 8002)
2. `python run.py` again without a `--stop` in between
3. `ps -eo pid,ppid,cmd | grep tmp.*\.sh` shows two wrappers; `cat logs/docker-control-service.pid` shows only the newer one
4. `python run.py --stop` then `ss -lptn 'sport = :8002'`: the port is still held (by the ghost's respawned child)
5. `python run.py` again: "port 8002 in use" warning, and the log starts interleaving

## Environment

- Linux, `python run.py` launcher on branch `dev` / current feature branches (the wrapper code has been like this since the `tt_setup` extraction in #850, with the restart loop added in #1187)
- Reproduces without hardware; this is purely host-process lifecycle

AnkittChauhan:
### Summary

- Resolves #1307 

where `docker-control-service` (:8002) and `inference-api` (:8001) left orphaned supervisor bash loops running in the background on `run.py --stop` or restarts. 

This PR prevents resurrection loops by discovering and terminating parent supervisor wrappers before child listeners, adds an automatic self-termination guard to the supervisor loops if their PID file is removed or overwritten, and ensures previous supervisors are strictly reaped before new ones start.

### Changes
- **Supervisor-first kill order (`tt_setup/services/_ports.py`)**: `kill_process_on_port` now climbs the process tree (`ps -o ppid=`) to find and terminate the parent supervisor wrapper *before* killing the child socket listener, stopping the 3-second respawn loop.
- **Ghost self-termination guard (`_docker_control.py`, `_fastapi.py`)**: The supervisor script checks its PID file before and after `sleep 3`. If the file was unlinked or reassigned to another process, it cleanly exits (`exit 0`).
- **Strict process reaping on start (`_docker_control.py`, `_fastapi.py`)**: Startup actively polls until any previous supervisor is confirmed dead before truncating the PID file or spawning a new instance. Added `pid <= 1` guards across kill helpers.

### Testing
- `python -m unittest discover tests` — all 425 tests pass.
- Added regression tests in `tests/test_services.py` for tree traversal, ghost self-termination, and reaping.
- Verified live on macOS: started detached service, deleted PID file, and called `kill_process_on_port(8002)`. Confirmed port 8002 stayed completely free with no supervisor respawning after 4+ seconds.

### DIFF
--- a/tt_setup/services/_ports.py
+++ b/tt_setup/services/_ports.py
@@ -330,7 +330,8 @@ def _kill_port_holder(port, no_sudo=False, quiet=False):
     if supervisor_pid and supervisor_pid != pid:
         if not quiet:
             print(f"🛑 Found parent supervisor wrapper with PID {supervisor_pid}. Stopping it first...")
-        _terminate_pid_graceful_then_force(supervisor_pid, use_sudo=use_sudo_for_kill, quiet=quiet)
+        if not _terminate_pid_graceful_then_force(supervisor_pid, use_sudo=use_sudo_for_kill, quiet=quiet):
+            return False
 
     if not quiet:
         print(f"🛑 Found process with PID {pid} using port {port}. Attempting to stop it...")
--- a/tt_setup/services/_ports.py
+++ b/tt_setup/services/_ports.py
@@ -170,6 +170,104 @@ def kill_process_on_port(port, no_sudo=False, quiet=False, attempts=3):
     return check_port_available(port)
 
 
+def _get_parent_pid(pid, use_sudo=False):
+    """Return parent PID of `pid`, or None."""
+    cmd = ["ps", "-o", "ppid=", "-p", str(pid)]
+    if use_sudo:
+        cmd.insert(0, "sudo")
+    result = run_command(cmd, check=False, capture_output=True)
+    if result.returncode == 0 and result.stdout.strip():
+        val = result.stdout.strip()
+        return int(val) if val.isdigit() else None
+    return None
+
+
+def _get_process_command(pid, use_sudo=False):
+    """Return command string of `pid`, or empty string."""
+    cmd = ["ps", "-o", "command=", "-p", str(pid)]
+    if use_sudo:
+        cmd.insert(0, "sudo")
+    result = run_command(cmd, check=False, capture_output=True)
+    if result.returncode == 0 and result.stdout.strip():
+        return result.stdout.strip()
+    return ""
+
+
+def _is_supervisor_wrapper(cmd_str):
+    """True if `cmd_str` matches a TT Studio bash supervisor wrapper."""
+    cmd = cmd_str.lower()
+    is_bash = "bash" in cmd or "sh" in cmd
+    has_script = ".sh" in cmd
+    targets_service = (
+        "docker-control-service" in cmd
+        or "inference-api" in cmd
+        or "docker_control" in cmd
+    )
+    if is_bash and (targets_service or (has_script and ("tmp" in cmd or "/var/folders" in cmd))):
+        return True
+    return False
+
+
+def _find_supervisor_wrapper_pid(pid, use_sudo=False, max_depth=5):
+    """Walk up the process tree to find any ancestor bash supervisor wrapper.
+
+    Handles dev mode where `uvicorn --reload` spawns intermediate reloader
+    processes between the bash wrapper and the socket listener.
+    """
+    curr = pid
+    for _ in range(max_depth):
+        ppid = _get_parent_pid(curr, use_sudo=use_sudo)
+        if not ppid or ppid <= 1:
+            break
+        cmd = _get_process_command(ppid, use_sudo=use_sudo)
+        if _is_supervisor_wrapper(cmd):
+            return ppid
+        curr = ppid
+    return None
+
+
+def _terminate_pid_graceful_then_force(pid, use_sudo=False, quiet=False):
+    """Send SIGTERM (-15), wait 2s, escalate to SIGKILL (-9) if still alive."""
+    try:
+        pid_int = int(pid)
+    except (ValueError, TypeError):
+        return True
+
+    if pid_int <= 1:
+        return True
+
+    pid_str = str(pid_int)
+    kill_cmd_graceful = ["kill", "-15", pid_str]
+    kill_cmd_force = ["kill", "-9", pid_str]
+    check_alive_cmd = ["kill", "-0", pid_str]
+
+    if use_sudo:
+        kill_cmd_graceful.insert(0, "sudo")
+        kill_cmd_force.insert(0, "sudo")
+        check_alive_cmd.insert(0, "sudo")
+
+    try:
+        run_command(kill_cmd_graceful, check=False, capture_output=True)
+        time.sleep(2)
+
+        result = run_command(check_alive_cmd, check=False, capture_output=True)
+        if result.returncode == 0:
+            if not quiet:
+                print(f"⚠️  Process {pid_str} still alive. Forcing termination...")
+            run_command(kill_cmd_force, check=True, capture_output=True)
+            if not quiet:
+                print(f"{C_GREEN}✅ Process {pid_str} terminated by force.{C_RESET}")
+        else:
+            if not quiet:
+                print(f"{C_GREEN}✅ Process {pid_str} terminated gracefully.{C_RESET}")
+    except Exception as e:
+        if not quiet:
+            print(f"{C_RED}⛔ Failed to kill process {pid_str}: {e}{C_RESET}")
+            print(f"{C_YELLOW}   You may need to stop it manually. Try: {' '.join(kill_cmd_force)}{C_RESET}")
+        return False
+    return True
+
+
 def _kill_port_holder(port, no_sudo=False, quiet=False):
     """One pass: find whoever holds `port` and stop it (see kill_process_on_port)."""
     pid = None
@@ -213,6 +311,9 @@ def _kill_port_holder(port, no_sudo=False, quiet=False):
             print(f"{C_YELLOW}⚠️  Could not find a specific process using port {port}. This is likely okay.{C_RESET}")
         return True
 
+    if str(pid).isdigit():
+        pid = int(pid)
+
     # NEVER kill Docker itself. On macOS/Docker Desktop the port is held by
     # com.docker.backend; killing it crashes the engine and the build then fails
     # with "Cannot connect to the Docker daemon". A TT Studio container holding
@@ -220,42 +321,18 @@ def _kill_port_holder(port, no_sudo=False, quiet=False):
     if _process_is_docker(pid):
         return "docker"
 
+    use_sudo_for_kill = not no_sudo and hasattr(os, "geteuid") and os.geteuid() != 0
+
+    # Stop supervisor wrapper first if one exists in the process tree.
+    # Killing only the socket holder leaves the supervisor loop alive to respawn
+    # its child 3 seconds later and recapture the port (Issue #1307).
+    supervisor_pid = _find_supervisor_wrapper_pid(pid, use_sudo=use_sudo_for_kill)
+    if supervisor_pid and supervisor_pid != pid:
+        if not quiet:
+            print(f"🛑 Found parent supervisor wrapper with PID {supervisor_pid}. Stopping it first...")
+        _terminate_pid_graceful_then_force(supervisor_pid, use_sudo=use_sudo_for_kill, quiet=quiet)
+
     if not quiet:
         print(f"🛑 Found process with PID {pid} using port {port}. Attempting to stop it...")
 
-    # Build kill commands
-    kill_cmd_graceful = ["kill", "-15", pid]
-    kill_cmd_force = ["kill", "-9", pid]
-    check_alive_cmd = ["kill", "-0", pid]
-    # os.geteuid() is POSIX-only; guard it so a non-POSIX host doesn't crash with
-    # AttributeError. (The kill/-15/-9 commands here are POSIX anyway.)
-    use_sudo_for_kill = not no_sudo and hasattr(os, "geteuid") and os.geteuid() != 0
-
-    if use_sudo_for_kill:
-        kill_cmd_graceful.insert(0, "sudo")
-        kill_cmd_force.insert(0, "sudo")
-        check_alive_cmd.insert(0, "sudo")
-
-    try:
-        run_command(kill_cmd_graceful, check=False, capture_output=True)
-        time.sleep(2)
-
-        result = run_command(check_alive_cmd, check=False, capture_output=True)
-        if result.returncode == 0:
-            if not quiet:
-                print(f"⚠️  Process {pid} still alive. Forcing termination...")
-            run_command(kill_cmd_force, check=True, capture_output=True)
-            if not quiet:
-                print(f"{C_GREEN}✅ Process {pid} terminated by force.{C_RESET}")
-        else:
-            if not quiet:
-                print(f"{C_GREEN}✅ Process {pid} terminated gracefully.{C_RESET}")
-
-    except Exception as e:
-        if not quiet:
-            print(f"{C_RED}⛔ Failed to kill process {pid}: {e}{C_RESET}")
-            print(f"{C_YELLOW}   You may need to stop it manually. Try: {' '.join(kill_cmd_force)}{C_RESET}")
-        return False
-
-    return True
-
+    return _terminate_pid_graceful_then_force(pid, use_sudo=use_sudo_for_kill, quiet=quiet)
--- a/tests/test_services.py
+++ b/tests/test_services.py
@@ -69,6 +69,12 @@ try:
 except ImportError:
     _dc_mod = M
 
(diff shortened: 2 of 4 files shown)


jashansinghTT:
Looks good @AnkittChauhan , once the license is updated from the CI I'll give it a go again.

tt_setup/services/_ports.py
```diff
@@ -241,6 +241,104 @@ def kill_process_on_port(port, no_sudo=False, quiet=False, attempts=3):
...
+    result = run_command(cmd, check=False, capture_output=True)
+    if result.returncode == 0 and result.stdout.strip():
+        return result.stdout.strip()
+    return ""
+
+
+def _is_supervisor_wrapper(cmd_str):
+    """True if `cmd_str` matches a TT Studio bash supervisor wrapper."""
+    cmd = cmd_str.lower()
+    is_bash = "bash" in cmd or "sh" in cmd
```
jashansinghTT: This match is the root of both problems below. `"sh" in cmd` is true for almost any path (`/home/jashan/...`), so in practice the check is: `.sh` and (`tmp` or `/var/folders`) anywhere in the cmdline, *or* one of the service dir names anywhere in the cmdline.

Concrete cases I ran through it:

```
True   /home/.../docker-control-service/.venv/bin/python3.12 .venv/bin/uvicorn api:app --reload   <- the reloader, not a wrapper
True   /bin/bash /tmp/ci_job_step.sh                                                               <- a user's own script
True   /bin/bash /var/folders/zb/T/my_launcher.sh                                                  <- same on macOS
True   /bin/bash -c source /tmp/session-snapshot.sh && eval 'ls /tmp'                              <- an IDE integrated-terminal shell
```

Our wrappers always look like `/bin/bash /tmp/tmpXXXX.sh <service-dir> <pid-file> .venv <log>`, so we can match that shape exactly and nothing else:

```suggestion
def _is_supervisor_wrapper(cmd_str):
    """True if `cmd_str` is one of *our* bash supervisor wrappers.

    The wrapper is always launched as `bash <tmpfile>.sh <service-dir> <pid-file> .venv <log>`,
    so require a bash interpreter, a .sh script, and one of our service dirs as an argument.
    Deliberately does not match `uvicorn --reload` (whose venv path also contains the service
    dir) or arbitrary user scripts living under /tmp.
    """
    parts = cmd_str.split()
    if len(parts) < 3 or not parts[0].endswith(("bash", "/sh", "sh")):
        return False
    if not parts[1].endswith(".sh"):
        return False
    return any(
        p.rstrip("/").endswith(("docker-control-service", "inference-api"))
        for p in parts[2:]
    )
```

tt_setup/services/_ports.py
```diff
@@ -241,6 +241,104 @@ def kill_process_on_port(port, no_sudo=False, quiet=False, attempts=3):
...
+    Handles dev mode where `uvicorn --reload` spawns intermediate reloader
+    processes between the bash wrapper and the socket listener.
+    """
+    curr = pid
+    for _ in range(max_depth):
+        ppid = _get_parent_pid(curr, use_sudo=use_sudo)
+        if not ppid or ppid <= 1:
+            break
+        cmd = _get_process_command(ppid, use_sudo=use_sudo)
+        if _is_supervisor_wrapper(cmd):
```
jashansinghTT, line 298: Repro for the false positive (any port, not just 8001/8002):

```
$ cat /tmp/ci_job_step.sh       # a user's launch script / CI step
#!/bin/bash
python3 -c 'import subprocess,time; subprocess.Popen(["bash","-c","python3 -m http.server 8765"]); time.sleep(600)'

$ setsid /tmp/ci_job_step.sh &
$ python -c 'from tt_setup.services._ports import kill_process_on_port; kill_process_on_port(8765)'
🛑 Found parent supervisor wrapper with PID 2586873. Stopping it first...
✅ Process 2586873 terminated gracefully.        <- that's /bin/bash /tmp/ci_job_step.sh, three levels up
🛑 Found process with PID 2586892 using port 8765. Attempting to stop it...
```

With the stricter match above this walk becomes safe, but I'd also return the *topmost* match rather than the first one, so a `--reload` tree (worker → reloader → bash wrapper) can never stop early:

```suggestion
def _find_supervisor_wrapper_pid(pid, use_sudo=False, max_depth=5):
    """Walk up the process tree and return the topmost ancestor that is one of our
    bash supervisor wrappers, or None.

    Handles dev mode where `uvicorn --reload` spawns intermediate reloader
    processes between the bash wrapper and the socket listener. Keeps climbing
    after a match so the walk can't stop at an intermediate process regardless
    of which pid lsof/ss reported for the port.
    """
    curr = pid
    found = None
    for _ in range(max_depth):
        ppid = _get_parent_pid(curr, use_sudo=use_sudo)
        if not ppid or ppid <= 1:
            break
        cmd = _get_process_command(ppid, use_sudo=use_sudo)
        if _is_supervisor_wrapper(cmd):
            found = ppid
        curr = ppid
    return found
```

tt_setup/services/_ports.py
```diff
@@ -284,49 +382,29 @@ def find_pid_with_command(base_cmd, use_sudo):
...
-                print(f"{C_GREEN}✅ Process {pid} terminated by force.{C_RESET}")
-        else:
-            if not quiet:
-                print(f"{C_GREEN}✅ Process {pid} terminated gracefully.{C_RESET}")
-
-    except Exception as e:
+    # Stop supervisor wrapper first if one exists in the process tree.
+    # Killing only the socket holder leaves the supervisor loop alive to respawn
+    # its child 3 seconds later and recapture the port (Issue #1307).
+    supervisor_pid = _find_supervisor_wrapper_pid(pid, use_sudo=use_sudo_for_kill)
```
jashansinghTT: This is where problem 2 bites. In `--dev` the socket is held by both the reloader and its worker:

```
$ ss -lptn 'sport = :8002'
users:(("python3.12",pid=2653579,fd=3),("uvicorn",pid=2653577,fd=3))   <- worker listed first
$ lsof -ti tcp:8002
2653577   <- reloader listed first
2653579
```

`find_pid_with_command` takes the first pid, so with `lsof` the walk starts at the reloader and finds the bash wrapper; with the `ss` fallback it starts at the worker, matches the reloader (its cmdline contains `docker-control-service`) and stops there. I verified the second path by hand: after killing what the walk returned plus the worker, the bash wrapper respawned uvicorn and 8002 was held again 6s later.

The two suggestions above (strict match + topmost ancestor) fix this regardless of ordering. Would be good to add a unit test where the reloader cmdline includes the service dir path, e.g. `/x/docker-control-service/.venv/bin/python .venv/bin/uvicorn api:app --reload`, since the current `test_find_supervisor_wrapper_pid_climbs_tree` uses a bare `python -m uvicorn --reload` and can't catch it.

tt_setup/services/_ports.py
```diff
@@ -241,6 +241,104 @@ def kill_process_on_port(port, no_sudo=False, quiet=False, attempts=3):
     return check_port_available(port)
 
 
+def _get_parent_pid(pid, use_sudo=False):
+    """Return parent PID of `pid`, or None."""
+    cmd = ["ps", "-o", "ppid=", "-p", str(pid)]
+    if use_sudo:
+        cmd.insert(0, "sudo")
```
jashansinghTT: nit: `ps -o ppid=` / `ps -o command=` don't need root to read other users' processes, so `use_sudo` here (and in `_get_process_command`) only adds sudo prompts on boxes without passwordless sudo, up to two per tree level. I'd drop the sudo path from these two helpers and keep it for the kill only.

tt_setup/services/_ports.py
```diff
@@ -241,6 +241,104 @@ def kill_process_on_port(port, no_sudo=False, quiet=False, attempts=3):
...
+    check_alive_cmd = ["kill", "-0", pid_str]
+
+    if use_sudo:
+        kill_cmd_graceful.insert(0, "sudo")
+        kill_cmd_force.insert(0, "sudo")
+        check_alive_cmd.insert(0, "sudo")
+
+    try:
+        run_command(kill_cmd_graceful, check=False, capture_output=True)
+        time.sleep(2)
```
jashansinghTT: nit, non-blocking: our wrapper's `shutdown()` trap waits up to 5s for uvicorn before exiting, so a fixed 2s here means almost every stop goes through the SIGKILL branch and prints "still alive. Forcing termination…" (I saw it on every run, and `--stop` reports ~4s for Docker control). Polling `kill -0` every 0.25s for up to ~6–8s, like `_wait_for_process_death` does in `_docker_control.py`, would let the graceful path actually succeed.

jashansinghTT (changes requested):
@AnkittChauhan thanks for taking this one on, the core of it is solid. I spent the afternoon on it on the p300 box: unit tests (485 green), the live repro from #1307, dev mode, `--stop` after a stale PID file, and a full `run.py` → `run.py` again → `--stop` cycle. The ghost guard, the strict reaping and the wrapper-first kill all do what they say. Nice work.

Two things in the new tree-walk in `_ports.py` need another pass before this goes in though, and they're both easy to hit in real life. Details inline, short version:

1. **`_is_supervisor_wrapper` is too broad and the walk goes 5 levels up from *any* port holder.** It matches anything with `sh` + `.sh` + `tmp` in the cmdline. I started a plain `python -m http.server` from a user script at `/tmp/ci_job_step.sh`, asked the launcher to free that port, and it killed the user's script three levels up. It also killed my own tool shell (which happened to be a `bash -c source .../snapshot-bash-….sh …` process). This runs on every start for 3000/8000/8080/8111 too, so it can reach processes that have nothing to do with TT Studio. On `dev` today that can't happen, so this is a regression.

2. **In `--dev` the result depends on which process `lsof`/`ss` lists first.** The uvicorn reloader's cmdline contains `docker-control-service` (via the venv path), so the heuristic matches it as well. Starting the walk from the reloader reaches the bash wrapper; starting from the spawned worker stops at the reloader, and after killing that the bash wrapper respawned uvicorn 6s later. It only passes end-to-end because `lsof` lists the reloader first. The `ss` fallback lists the worker first, so on a box without `lsof` the dev-mode ghost comes straight back.

Suggested fix for both: match only our own wrapper shape (bash running a `.sh` whose arguments include the `docker-control-service` or `inference-api` dir), drop the generic tmp/`/var/folders` fallback, and keep climbing to the topmost match instead of returning the first one. I've put concrete snippets inline.

Two nits, non-blocking: the fixed 2s grace escalates to SIGKILL on basically every stop (the wrapper's own shutdown handler waits up to 5s for uvicorn), and `ps` doesn't need `sudo` to read ppid/cmd.

Happy to re-test as soon as it's updated, the harness is still on the box.

AnkittChauhan:
Hey @jashansinghTT , really appreciate for the detailed review and testing. all feedback points have been addressed, tested, and pushed.

jashansinghTT (approved):
Thanks Ankit, this looks good now. All the points from the previous review are addressed, and I re-tested the branch on the P300 box: unit tests, the live repro from #1307, dev mode, --stop with stale PID files and a full restart cycle. Everything holds. One small follow-up on the kill -0 zombie check will come as a separate PR.

PR #1318, opened 2026-09-09
fix(frontend): handle insecure origins for microphone

### LINKED ISSUE #1313: Speech to Text: microphone button silently does nothing on non-secure origins (http://<ip>:3000)
**Describe the bug**

On the Speech to Text page, clicking the microphone button does nothing when the browser does not expose `navigator.mediaDevices`. No toast, no alert, no console message. The button simply appears dead.

The common way to hit this is opening TT-Studio from another machine over plain HTTP, e.g. `http://<box-ip>:3000`. Browsers only expose the microphone API on secure origins (`https://` or `localhost`), so on a LAN IP or hostname the API is absent and recording is impossible. Nothing in the UI tells the user this.

The recorder checks for the API and only has a happy path:

https://github.com/tenstorrent/tt-studio/blob/dev/app/frontend/src/components/speechToText/AudioRecorderWithVisualizer.tsx#L148

```ts
if (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
  navigator.mediaDevices.getUserMedia({ audio: true }).then(...).catch((error) => {
    alert("Microphone access error: " + error.message);
  });
}
// no else branch
```

A related inconsistency: when the API exists but permission is denied, the error is surfaced with a browser `alert()` rather than the standard toast used elsewhere in the app.

**To Reproduce**

1. Open TT-Studio from a different machine using the box's IP or hostname over HTTP, e.g. `http://192.168.x.y:3000`
2. Models Deployed → a running speech model (e.g. distil-large-v3) → open Speech to Text
3. Click the microphone button
4. Nothing happens. Backend receives no request. Browser console is silent.

To confirm the cause, run this in the browser console on the page:

```js
[window.isSecureContext, !!navigator.mediaDevices]
```

Either value `false` reproduces the dead button.

The same flow works on `http://localhost:3000` (e.g. via an SSH tunnel). Verified on a P300x2 box against distil-large-v3: record → stop → send → 200 with a correct transcript.

**Expected behavior**

- When the microphone API is unavailable, show a toast explaining that microphone access needs a secure origin, and suggest using `localhost` (SSH tunnel) or HTTPS. Ideally disable the mic button with a tooltip instead of leaving it clickable.
- When permission is denied, use the standard toast instead of `alert()`.
- Optionally, keep the upload panel (from #1282) as the visible alternative in that state, since file upload does not need the microphone API.

**Desktop**

- OS: Linux (TT-Studio host), client browser on a separate machine
- Browser: Chrome
- Version: any current version; this is standard secure-context behaviour, not a browser bug

**Additional context**

Found while testing #1282. The Voice Agent page relies on the same API and likely has the same silent failure over a non-secure origin; worth checking at the same time.

AnkittChauhan:
## Description
On plain HTTP non-secure origins (e.g. `http://<ip>:3000`), modern browsers do not expose `navigator.mediaDevices`. Previously, clicking the microphone button silently failed with no visual feedback, leaving users to think the button was dead. Additionally, permission denial triggered a disruptive native `window.alert()`.
- Fixes #1313.

## Changes
- **Shared capability check**: Added `isAudioRecordingSupported()` and `getMicrophoneErrorMessage()` in `src/lib/mediaUtils.ts`.
- **Speech to Text**:
  - Pre-checks capability in `startRecording()` and surfaces `customToast.warning` explaining the HTTPS / SSH tunnel requirement.
  - Replaced native `alert()` with `customToast.error()` providing clear browser permission instructions.
  - Visually disables the mic button with a Radix UI tooltip.
  - Keeps the file upload panel prominently visible as the working alternative.
- **Voice Agent**:
  - Eliminated false start chime on insecure origins.
  - Disabled mic button with tooltip title and toast warning on interaction.
  - Guarded `useWakeWord` hook against unhandled `TypeError` exceptions.

## How Has This Been Tested?
- **Manual / Browser Testing**:
  - Simulated insecure origin (`window.isSecureContext = false` / `navigator.mediaDevices = undefined`): verified mic button is disabled with tooltip, warning toast fires on interaction, and file upload card remains visible with "Available on HTTP" badge.
  - Tested on both Speech to Text (`/speech-to-text`) and Voice Agent (`/voice-agent`).
  - Simulated microphone permission denied: verified `customToast.error` displays settings guidance with zero blocking `window.alert()` dialogs.
- **Lint & Headers**: Passed `npx eslint --quiet` and `npm run header:check:changed` with zero errors.
- **Build**: Production bundle compiles cleanly via `npx vite build`.

### DIFF
--- a/app/frontend/src/components/speechToText/AudioRecorderWithVisualizer.tsx
+++ b/app/frontend/src/components/speechToText/AudioRecorderWithVisualizer.tsx
@@ -20,6 +20,13 @@ import {
   TooltipProvider,
   TooltipTrigger,
 } from "../ui/tooltip";
+import { customToast } from "../CustomToaster";
+import {
+  isAudioRecordingSupported,
+  INSECURE_CONTEXT_MIC_MESSAGE,
+  INSECURE_CONTEXT_TOOLTIP_MESSAGE,
+  getMicrophoneErrorMessage,
+} from "../../lib/mediaUtils";
 
 type Props = {
   className?: string;
@@ -50,6 +57,7 @@ export const AudioRecorderWithVisualizer = ({
   const [audioUrl, setAudioUrl] = useState<string | null>(null);
   const [hasRecordedBefore, setHasRecordedBefore] = useState<boolean>(false);
   const [isPlaying, setIsPlaying] = useState<boolean>(false);
+  const isMicSupported = useMemo(() => isAudioRecordingSupported(), []);
 
   // Calculate the hours, minutes, and seconds from the timer
   // const hours = Math.floor(timer / 3600);
@@ -128,6 +136,11 @@ export const AudioRecorderWithVisualizer = ({
   }
 
   function startRecording() {
+    if (!isAudioRecordingSupported()) {
+      customToast.warning(INSECURE_CONTEXT_MIC_MESSAGE);
+      return;
+    }
+
     if (mediaRecorderRef.current.stream) {
       mediaRecorderRef.current.stream
         .getTracks()
@@ -194,8 +207,8 @@ export const AudioRecorderWithVisualizer = ({
           };
         })
         .catch((error) => {
-          alert("Microphone access error: " + error.message);
           console.error("Microphone access error:", error);
+          customToast.error(getMicrophoneErrorMessage(error));
         });
     }
   }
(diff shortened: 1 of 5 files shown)


jashansinghTT (approved):
LGTM

PR #1319, opened 2026-09-10
feat: embedding models on fewer chips and dedicated interface to use them

rnabeelTT:
## Summary

Adds a first-class embedding model experience to TT-Studio and cleans up how model availability/deploy exceptions are maintained.

**User-facing:**
- Embedding models (bge-m3, Qwen3-Embedding-0.6B, Qwen3-Embedding-4B) are now deployable from the catalog — previously hidden/no UI.
- New Embedding Demo page: single-text embedding and document-upload/search, matching the design language of the other model demo pages.
- RAG Management can now embed documents using a deployed embedding model instead of only the default local model.
- Advanced deploy config for embedding models offers a real throughput/cost tradeoff: 1 device (default, cheapest), 2 devices (one P300 card, ~2x throughput), or all 4 devices (~4x throughput), instead of only an all-or-nothing board allocation.

**Developer-facing:**
- All model-availability exceptions (known-broken models/devices), mesh device fallbacks, deploy-time image/trace-region overrides, and the new chip-tier definitions now live in one reviewable `model_overrides.toml`, validated with clear errors (unknown reason, typo'd device, duplicate entry) — mirroring tt-cli's `model_support_overrides.toml` pattern instead of several scattered Python dicts across two files.
- The three embedding models' single-chip/dual-chip runtime specs are generated automatically from the live tt-inference-server artifact on every catalog sync, instead of being hand-maintained files that could drift from upstream.

## Test plan

- Run `python run.py` (or `python run.py --dev` for hot reload) and open the TT-Studio UI.
- From the model catalog, deploy `bge-m3` or `Qwen3-Embedding-0.6B` with default (1-device) settings and confirm it reaches Healthy.
- Open the new Embedding Demo page for the deployed model: submit a single text and confirm an embedding vector comes back, then upload a document and confirm search/retrieval works.
- Go to RAG Management, add/select a collection, and confirm the deployed embedding model appears as an option and successfully embeds an uploaded document with it.
- Deploy an embedding model again with Advanced Config open on a P300x2 board: confirm "1 Device", "2 Devices", and "All Devices" are all selectable, each deploys successfully, and 1 Device is pre-selected by default.
- Stop all deployments and confirm chip slots reset cleanly between the three tiers.
- Optionally: edit an entry in `app/backend/shared_config/model_overrides.toml` (e.g. change a `verified_on` or add a bogus device name) and re-run the sync script to confirm validation errors surface clearly instead of failing silently.
- Go to the apps marketplace, ensure OpenWebUI and AnythingLLM display a dropdown to select an embedding model from the one deployed and their native one. Launch one and confirm in settings that it is set up. Upload a document to test end to end.
- Deploy an STT or TTS model and confirm those too show up in the app card drop-downs, confirm that they work end to end.

### DIFF
--- a/app/backend/docker_control/docker_utils.py
+++ b/app/backend/docker_control/docker_utils.py
@@ -465,11 +465,22 @@ def deploys_whole_board(impl, board_type=None):
     return infer_inference_server_device(impl, board_type) not in _SINGLE_CHIP_DEVICE_NAMES
 
 
-def run_container(impl, weights_id, device_id=0, host_port=None, use_image_override=True):
+def run_container(impl, weights_id, device_id=0, host_port=None, use_image_override=True, force_full_board=False):
     """Run a docker container.
 
     For FACE_RECOGNITION model type, uses docker-control-service directly.
     For all other model types, uses TT Inference Server API.
+
+    force_full_board: opt-in whole-board mesh deploy for a model whose default is
+    single-chip via a runtime_model_spec_overrides entry (see
+    shared_config.model_config.ModelImpl and
+    sync_models_from_inference_server.STUDIO_CHIP_TIER_MODELS), but which
+    also has a genuine, already-working whole-board spec upstream (e.g.
+    Qwen3-Embedding-0.6B/4B and bge-m3 on P300x2, at 4x throughput). Set by
+    docker_control.views.DeployView from the request's force_full_board flag,
+    gated to models it actually applies to. Bypasses the single-chip override
+    entirely: device resolves to the board mesh name, so runtime_model_spec_json
+    is never set below and normal (already-working) spec resolution runs instead.
     """
     # Face recognition bypasses TT Inference Server — deploy via docker-control-service
     if impl.model_type == ModelTypes.FACE_RECOGNITION:
@@ -485,11 +496,30 @@ def run_container(impl, weights_id, device_id=0, host_port=None, use_image_overr
         # ("t3k"). We use chips_required + board_type to pick the right name.
         from shared_config.model_config import infer_chips_required
         board_type = detect_board_type()
-        chips_required = infer_chips_required(impl.device_configurations)
-        device = infer_inference_server_device(impl, board_type)
+        if force_full_board:
+            chips_required = 4
+            device = map_board_type_to_device_name(board_type)
+        else:
+            chips_required = infer_chips_required(impl.device_configurations)
+            device = infer_inference_server_device(impl, board_type)
+            # Card-pair promotion: a model with a P300 (2-chip) override spec that's
+            # asked for an explicit 2-slot device_id forming one physical P300 card
+            # ((0,1) or (2,3)) runs as that one card, not two independent p150 chips.
+            # Gated on the override existing so this never touches ordinary
+            # single-chip models -- see STUDIO_CHIP_TIER_MODELS.
+            requested_ids = [x.strip() for x in str(device_id).split(",") if x.strip() != ""]
+            if (
+                device == "p150"
+                and board_type == "P300x2"
+                and len(requested_ids) == 2
+                and sorted(int(x) for x in requested_ids) in ([0, 1], [2, 3])
+                and "p300" in {k.lower() for k in (impl.runtime_model_spec_overrides or {})}
+            ):
+                device = "p300"
+                chips_required = 2
         logger.info(
             f"Device name '{device}' for {impl.model_name} "
-            f"(board={board_type}, chips_required={chips_required})"
+            f"(board={board_type}, chips_required={chips_required}, force_full_board={force_full_board})"
         )
 
         BASE_SERVICE_PORT = 7000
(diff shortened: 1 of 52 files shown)


jashansinghTT (changes requested):
Requesting changes: bge-m3 deploys with expects_weights=False, but DeploymentProgress.tsx still hardcodes the 0–25% pull band, so the deploy card shows 18% while the Downloads tray shows 53% for the same pull. Please route the card through compactPercent (lib/deployProgress.ts) so both agree.

rnabeelTT:
@jashansinghTT thanks for pointing this out, I have resolved it

app/backend/shared_config/models_from_inference_server.json
```diff
@@ -1132,14 +1133,9 @@
...
-      "available_in_studio": false,
-      "unavailable_reason": "known_broken",
-      "unavailable_details": "No supported device left: P300x2: Non-functional deploy (#869). Separate from the embedding modality gap: this model does not run. Verified broken on a P300x2 (4x p300c Blackhole) box. Untested on this model's other boards, so it stays available there.",
-      "unavailable_devices": {
-        "P300x2": {
-          "reason": "known_broken",
-          "details": "Non-functional deploy (#869). Separate from the embedding modality gap: this model does not run. Verified broken on a P300x2 (4x p300c Blackhole) box. Untested on this model's other boards, so it stays available there."
-        }
+      "runtime_model_spec_overrides": {
+        "P150": "app/backend/shared_config/runtime_model_specs/bge-m3-p150.json",
```
jashansinghTT, line 1170: These paths are committed, but the files they reference are not: `runtime_model_specs/*.json` is caught by the root `.gitignore` `*.json` rule, and the only thing that writes them is `apply_chip_tier_overrides` during a catalog sync that `run.py` skips whenever this JSON already exists. On a fresh checkout of this branch the P150/P300 rows are offered in the UI but every deploy fails with `runtime_model_spec_json does not exist: .../bge-m3-p150.json`. Same for the Qwen3-Embedding-0.6B / 4B entries below.

Either commit the six generated spec files (add `!app/backend/shared_config/runtime_model_specs/*.json` to `.gitignore`), or don't ship these paths in the catalog and let the sync add them at runtime.

app/backend/shared_config/sync_models_from_inference_server.py
```diff
@@ -70,192 +73,138 @@
...
+        devices = list(model.get("device_configurations") or [])
+        specs = dict(model.get("runtime_model_spec_overrides") or {})
+        for device_type, device_ids in STUDIO_CHIP_TIERS[model_name].items():
+            spec_path = RUNTIME_MODEL_SPECS_DIR / f"{model_name.lower()}-{device_type.lower()}.json"
+            if base is not None:
+                derived = _derive_chip_tier_spec(base, device_type, device_ids)
+                with open(spec_path, "w") as f:
+                    json.dump(derived, f, indent=2)
+                    f.write("\n")
+            elif not spec_path.exists():
```
jashansinghTT, line 198: When `base` is None and the file doesn't exist, this `continue` skips writing the file but `specs` and `devices` were seeded from the existing catalog entry a few lines up, so a stale `runtime_model_spec_overrides[P150]` path and the `P150` device both survive into the written catalog. That's exactly the state that produces the `runtime_model_spec_json does not exist` deploy error. In this branch the tier should be dropped here instead:

```python
elif not spec_path.exists():
    specs.pop(device_type, None)
    if device_type in devices:
        devices.remove(device_type)
    continue
```

app/backend/docker_control/docker_utils.py
```diff
@@ -579,12 +576,29 @@ def run_container(impl, weights_id, device_id=0, host_port=None, use_image_overr
...
+        # hand-authored spec file: run.py's own spec resolution hard-rejects an
+        # undeclared (model, device) pair before any override flag ever runs, and
+        # --runtime-model-spec-json is the documented bypass -- it's used as-is, so
+        # override_tt_config below would be silently ignored and is skipped.
+        spec_overrides = {
+            k.lower(): v for k, v in (impl.runtime_model_spec_overrides or {}).items()
+        }
+        spec_path = spec_overrides.get(device)
+        if spec_path:
+            payload["runtime_model_spec_json"] = str(
```
jashansinghTT, line 591: This hands the inference server a host path without checking it exists, so the failure surfaces as an opaque run.py error after the deploy has already been accepted and a chip slot reserved. Worth failing fast here with an actionable message, e.g.

```python
spec_file = Path(backend_config.host_tt_studio_root) / spec_path
if not spec_file.exists():
    raise RuntimeError(
        f"Missing runtime model spec for {impl.model_name} on {device}: {spec_file}. "
        "Run `python run.py --resync` to regenerate it."
    )
```

(If the backend container doesn't mount the repo root, check via the docker-control-service instead, or at least include the `--resync` hint in the error surfaced to the UI.)

jashansinghTT (changes requested):
Tested on this branch on a P300x2 box: deploying **bge-m3 on P150** fails immediately with

```
Deployment error: runtime_model_spec_json does not exist:
/home/jashan/tt-studio/app/backend/shared_config/runtime_model_specs/bge-m3-p150.json
```

Root cause: the committed catalog points the P150/P300 tiers at spec files under `app/backend/shared_config/runtime_model_specs/`, but those files are (a) gitignored by the root `*.json` rule and (b) only produced by `apply_chip_tier_overrides` inside the catalog sync. `run.py` only runs the sync when the catalog JSON is missing (or with `--resync` / `--reconfigure-inference-server` / `--pull-branch`), and the catalog JSON is committed, so on any checkout of this branch the sync never runs and the referenced spec files never exist. The UI still offers P150/P300 because `device_configurations` already lists them, so every single-chip embedding deploy fails out of the box.

Confirmed by running the sync manually (`python app/backend/shared_config/sync_models_from_inference_server.py`), which generated all six files and made the deploy path reachable.

Suggested fix (pick one, first is simplest and makes fresh checkouts work):
1. Commit the generated spec files and un-ignore the directory (`!app/backend/shared_config/runtime_model_specs/*.json` in `.gitignore`). They're small and the sync already regenerates them from the live artifact, so drift is self-healing.
2. Or make the launcher's `should_sync` in `tt_setup/cli/_run.py` also fire when any path referenced by `runtime_model_spec_overrides` is missing on disk.

Plus the two smaller inline points below.

jashansinghTT (approved):
Looks good to me.

PR #1323, opened 2026-09-15
fix: make training model send p150 to match inference server's expectation

mcupacTT:
- training needs to send --device p150 to match tt-inference-server expectation
- training docker image is corrected in `models_from_inference_server.json`


jashansinghTT (approved):
LGTM

PR #1324, opened 2026-09-15
feat: add full support of custom dataset for training

mcupacTT:
Wires custom datasets end-to-end so a user-uploaded file is actually trained on, instead of silently falling back to the built-in SST2 recipe. Users can upload a `.json` or `.jsonl` dataset, preview it (with sampling for very large files), pick a prompt template, and map their columns to the template's fields. On submit, the backend validates and normalizes the file, stages it into the running training container's volume, and rewrites the request into the training server's custom-dataset contract (`train_dataset_path` / `file_type` / `template` / `column_mapping`). The hyperparameter inputs are now editable as well.


mcupacTT:
copilot changes should be resolved now

jashansinghTT (approved):
LGTM, the custom datasets work at the present

PR #1326, opened 2026-09-16
feat: enable template insertion when chatting to fine-tuned model

mcupacTT:
Fine-tuned models are trained on a fixed instruction template (e.g. Alpaca) and need to be tested in that exact format, not through the standard chat interface. This adds a dedicated "template mode" to the chat page for those models.

When a model deployed with merged fine-tuned weights is selected, the composer switches to a template panel where you define your training template with `{placeholder}` fields, fill them in, and send the rendered prompt straight to the model's raw `/v1/completions` endpoint (bypassing chat-template wrapping). A Template/Chat toggle lets you fall back to the normal chat composer at any time.

**Highlights:**
- Detects fine-tuned deployments via `host_weights_dir` (persisted at deploy time and surfaced through the deployed-models API).
- Template composer with editable prompt template, auto-generated input fields, a live preview of the exact string sent, and an editable stop-sequences field.
- Defaults to deterministic decoding (temperature 0) on entering template mode; hides chat-only UI (system prompt, welcome examples) that doesn't apply.
- Backend routes requests carrying a raw `prompt` to `/v1/completions` instead of chat-completions.

---

Here is how the feature appears in the UI:
image


mcupacTT:
copilot comments should be resolved now

jashansinghTT (approved):
LGTM

Issue #1329, opened 2026-09-16
Chip slot allocator ignores chips held by externally launched containers; deploy blocks on device lock, then dies on stale sysmem and is never marked failed


jashansinghTT:
## Summary

The chip slot allocator only counts TT-Studio's own `ModelDeployment` rows when deciding which chips are free. A container launched outside TT-Studio (here: the `tt-model` CLI's Devstral vLLM server, which mapped the whole `/dev/tenstorrent` directory and set `TT_METAL_VISIBLE_DEVICES=0,1,2,3`) held every chip on a P300x2 box for two days while `/docker/chip-status/` reported all four slots as available. Deploying bge-m3 into "available" slot 0 then:

1. blocked for 22 minutes inside the container on the UMD `CHIP_IN_USE_0_PCIe` robust mutex, with the UI stuck on "Starting" and no indication why;
2. when the other container was stopped and the lock released, died immediately with `Sysmem mapped at unexpected NOC address (likely a stale process holding sysmem)` because the chip had not been reset;
3. was never marked failed by the backend, since the forge worker does not retry and the health probe kept returning 405 "not ready" forever.

Recovery required stopping the external container by hand, stopping bge-m3 through TT-Studio (which runs `tt-smi -r 0`), and redeploying.

## Environment

- Host: p300c, P300x2 board (4 chips), KMD 2.8.0, firmware bundle 19.13.1
- TT-Studio dev stack at `88844652` on `raahem/embedding-interface` (allocator code is unchanged on `dev`)
- bge-m3 via `ghcr.io/tenstorrent/tt-media-inference-server-forge:0.20.0-de59f8a`, catalog id `id_tt-metal-bge-m3-v0.20.0`
- External container: `tt-model-devstral-small-2-24b-instruct-2512-p150x4`, labels `org.tenstorrent.tt-model.kind=vllm-plugin`, `org.tenstorrent.tt-model.profile=p150x4`, devices `/dev/tenstorrent` (whole directory), `ipc=host`, env `MESH_DEVICE=P150x4`, `TT_METAL_VISIBLE_DEVICES=0,1,2,3`

## Timeline (2026-09-16, UTC)

| Time | Event |
|---|---|
| 09-14 13:39 | Devstral container started by `tt-model` CLI, owns chips 0-3, serves on :20000 |
| 18:01:03 | bge-m3 deployed from the UI, allocator hands out slot 0 |
[...]

PR #1330, opened 2026-09-16
fix(backend): show chips held by other software as occupied and block deploys onto them

jashansinghTT:
Chips held by a model that some other tool started (tt-model CLI, tt-inference-server run.py, a plain docker run) now show as in use on the deploy page, and TT-Studio refuses to deploy onto them instead of hanging on the driver's chip lock. Fixes the occupancy half of #1329.

How to test it, no hardware model needed:

1. Start TT-Studio with `python run.py --dev` and open http://localhost:3000 in your browser.
2. Go to the model deploy page. Note which devices show IDLE.
3. In a terminal, pretend some other software grabbed chip 2:
   `docker run -d --rm --name fake-holder --device /dev/tenstorrent/2 alpine sleep 600`
4. Back in the browser, within a few seconds Device 02 flips to IN USE with the name `fake-holder` under it and no STOP button, since TT-Studio does not own it.
5. Pick any single-chip model, open the advanced device picker, and choose device 2. The deploy is refused with "Chip slot 2 is occupied by fake-holder (container not deployed by TT-Studio)." Leave the device on automatic and the model lands on a free chip instead of chip 2.
6. Run `docker stop fake-holder`. Device 02 goes back to IDLE on the next refresh.
7. Sanity check: the TT-Studio backend container itself also binds /dev/tenstorrent, and it must not appear as a holder. All devices read IDLE with nothing else running.

The same thing from a terminal, if you prefer curl:

```
curl -s localhost:8000/docker/chip-status/
curl -s -X POST localhost:8000/docker/deploy/ -H 'Content-Type: application/json' \
  -d '{"model_id":"id_tt-metal-distil-large-v3-v0.17.0","device_id":"2"}'
```

- [x] Move the /dev/tenstorrent mount detector and infra-container prefixes from views.py into chip_allocator.py (pure move, first commit)
- [x] Allocator scans running containers for bound /dev/tenstorrent nodes and merges them into chip-status, occupancy, and the 409 messages
- [x] Skips TT-Studio's own containers and containers already tracked as deployments; falls back to records alone if Docker cannot be listed
- [x] Six new allocator tests; `manage.py test docker_control.tests docker_control.test_deploy_reservation` runs 68, the only two failures are pre-existing on dev (`test_allocator_rejects_while_a_full_board_model_is_starting`, `test_placeholder_reserves_the_slot_and_reads_as_pending`)
- [x] Verified live on a P300x2 box against the real docker-control service: fake-holder on chip 2 shows occupied with its name, manual deploys onto chip 2 and onto a chip held by a running bge-m3 container both return 409 naming the container, slot frees when the container stops, backend container is not counted

Not in this PR, still open on #1329: surfacing the "waiting for chip lock" log line as a deploy phase, marking a deploy failed when the inference worker dies but /health keeps returning 405, and narrowing a whole-directory bind by TT_METAL_VISIBLE_DEVICES (whole directory is treated as all chips).


rnabeelTT (approved):
Tested by deploying bge-m3 and Llama3.1-8b-instruct from other instances before launching studio. This branch correctly detected that those chips were already occupied, and correctly restricted deployments to them.

LGTM

PR #1341, opened 2026-09-18
fix(frontend): keep advanced device selection from resetting on every…

rnabeelTT:
## Summary 
The advanced device configuration ignored clicks: picking a device or an "All Devices" tier left the selection empty and Deploy stuck on "Select a Device", for Llama-3.1-8B card pairs, single-device models like speecht5_tts and the embedding models' 1/2/4-device tiers.

That's fixed now.


jashansinghTT (approved):
LGTM

PR #1338, opened 2026-09-17
Rc v2.11.0

rnabeelTT:
## Rc v2.11.0

Cut from `main`; validated changes cherry-picked from `dev` (see CONTRIBUTING → Release Process).

### Test plan

- [ ] Fresh `python run.py` completes (setup → ready panel)
- [ ] Backend healthy: `curl localhost:8000/up/` and `curl localhost:8000/models/health/`
- [ ] Frontend loads at `localhost:3000`
- [ ] Deploy a model end-to-end; inference server healthy: `curl localhost:8001/health`
- [ ] `python run.py --stop`, then a restart works
- [ ] `IS_QB2` is off in `.env.default` (rc-* branches never ship it on)
- [ ] Release notes drafted and applied to this description

PR #1344, opened 2026-09-21
fix(training): stop granting world access on custom dataset dirs and staged copies

jashansinghTT:
Cycode flags several permissive chmod calls on the v2.11.0 RC (#1338). This clears the ones that can be tightened without breaking the shared training volume.

- [x] Custom-datasets upload dir is owner-only (`stat.S_IRWXU`) instead of world-writable sticky; only the backend reads or writes it
- [x] Datasets staged into `volume_id_*/custom_datasets` are chowned to the training container uid (1000) with owner-only modes instead of `0o755`/`0o644`
- [x] Store test spells its owner-only cleanup modes with `stat` constants
- [x] Verified live in the dev stack: new dir lands as `0700 root`, staged copy as `0700`/`0600` owned by uid 1000, and tt-inference-server's `check_path_permissions_for_uid(..., 1000, need_write=True)` passes on the result; `/up/` 200

Left as-is on purpose: the `training_volume` root (`docker_utils.py`) must stay writable by the host user running run.py and pass tt-inference-server's uid-1000 read/write check, and `_chmod_merged_dir_readable` must let the host-side inference server read weights written by uid 1000. Both need group/other bits, which this rule rejects outright, so they need a design change rather than a mode tweak.

Note: `test_falls_back_to_memory_when_volume_unwritable` fails identically on `dev` (as root the dir is never unwritable; on the host the fallback returns None), so it is pre-existing.


rnabeelTT (approved):
LGTM

